PR #26, opened 2020-02-11
Fix issue with Starlette mount

Prillan:
This PR fixes an issue where Starlette mounts weren't handled properly. I'm not sure if they were supported to begin with, but at least it doesn't crash on them now (and everything works the same as before).

### DIFF
--- a/spectree/plugins/starlette_plugin.py
+++ b/spectree/plugins/starlette_plugin.py
@@ -78,7 +78,7 @@ class StarlettePlugin(BasePlugin):
                 if route.path.startswith(f'/{self.config.PATH}'):
                     continue
 
-                func = route.endpoint
+                func = route.app
                 if isinstance(func, partial):
                     try:
                         func = func.__wrapped__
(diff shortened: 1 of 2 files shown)


spectree/plugins/starlette_plugin.py
```diff
@@ -71,13 +71,18 @@ def register_route(self, app):
...
         def parse_route(app, prefix=''):
             for route in app.routes:
                 if route.path.startswith(f'/{self.config.PATH}'):
                     continue
 
+                if isinstance(route, Mount):
+                    parse_route(route, prefix=f'{prefix}{route.path}')
+                    continue
+
                 func = route.endpoint
```
kemingy:
```suggestion
                func = route.app
```
Sorry, my bad. This should fix this problem. No need to check if it's an instance of `Mount`

Prillan: Cool, I've pushed this change instead.

PR #41, opened 2020-03-15
Dev: make it easy to use customized backend

kemingy:


spectree/plugins/flask_plugin.py
```diff
@@ -92,19 +92,23 @@ def parse_path(self, route):
...
+
     def validate(self, func, query, json, headers, cookies, resp, *args, **kwargs):
         from flask import request, abort, make_response, jsonify
 
         try:
-            arg = request.args or {}
-            data = request.get_json() or {}
-            head = request.headers or {}
-            cook = request.cookies or {}
+            arg, json_data, header, cookie = self.get_req_data(request)
```
yedpodtrzitko: In order to do the arguments type checking inside this method, it also needs `query` (where the validation model is) as a parameter, so it can compare the request data with field types.

kemingy: Yes. To make it more general, I think it's better to pass `(query, headers, cookies, json)` these 4 models.

Then, let's just write a function to handle all the request validation part.

yedpodtrzitko: lgtm

PR #43, opened 2020-03-16
add helper for extracting request data

yedpodtrzitko:
Hi,
I've added the helper `get_request_data` into all three backends to make their interface identical. It has the `query` parameter, which is unused as it is, but it allows customization.

Backend passing functionality isn't included, as it feels like a separate concern which should be in its own MR/commit.


kemingy:
Please make sure it passes all the tests.

BTW, please help check if the current #41 fulfills your needs.

yedpodtrzitko:
#41 looks alright, closing this one then.

PR #46, opened 2020-04-13
Falcon ASGI support

### LINKED ISSUE #45: Async Falcon support?
I'm currently experimenting with Falcon 3.0.0a1 and was wondering if you have plans to add an async Falcon plugin for SpecTree (3.0 will support both WSGI and ASGI applications)?

If not, this might be something I would be interesting in contributing since I really like that SpecTree provides a way to both handle input/output (de)serialization with Pydantic *and* autogenerates an OpenAPI document similar to FastAPI. What would your preferred method of implementing this? A completely separate plugin, a subclass of the existing Falcon plugin, something else?

onecrayon:
Closes #45.

This is a work-in-progress (in particular, I have not implemented any tests), but I wanted to solicit feedback on my initial working approach. Main changes:

* Backend plug-in specifies whether it is async or not with `IS_ASYNC` class constant (defaults to `False` on BasePlugin)
* Added light child class for FalconPlugin that switches to coroutines for methods that must be coroutines for Falcon to work
* Added light child classes for the view classes, since they also need to be asynchronous

The reason I didn't write any tests was that since Falcon 3.0 is currently in pre-release, I was leery of testing against it. I'm honestly not too sure how to handle this; maybe this PR doesn't really make sense to merge in until Falcon 3.0 is released. ¯\_(ツ)_/¯

Thoughts?


spectree/plugins/falcon_plugin.py
```diff
@@ -24,10 +24,25 @@ def on_get(self, req, resp):
...
+
+
+DOC_CLASS = [x.__name__ for x in (DocPage, OpenAPI, DocPageAsgi, OpenAPIAsgi)]
+
+HTTP_422 = '422 Unprocessable Entity'
+
+
+class FalconPlugin(BasePlugin):
+    BASE_ROUTE = OpenAPI
+    BASE_DOC_ROUTE = DocPage
```
kemingy, line 44: `BASE_ROUTE` and `BASE_DOC_ROUTE` seem not very explicit?

kemingy:
Thanks for your contribution. The total design looks good to me.

I think you can work on the test part after Falcon's stable release.

onecrayon:
> I think you can work on the test part after Falcon's stable release.

Sounds good; I'll maintain this PR as I test it locally against Falcon's pre-release content, and ping you when 3.0 is released and I can get some tests written against a stable version.

kemingy:
Falcon 3.0 https://github.com/falconry/falcon/releases/tag/3.0.0 is released. We can continue this work.

onecrayon:
@kemingy I finally got a moment to myself, and I've updated this PR to bring it into line with the latest Falcon 3 WSGI support, and added tests. A lot of it is just copy/paste building on my previous code, though. 😅

spectree/plugins/falcon_plugin.py
```diff
@@ -192,3 +214,64 @@ def bypass(self, func, method):
...
+    OPEN_API_ROUTE_CLASS = OpenAPIAsgi
+    DOC_PAGE_ROUTE_CLASS = DocPageAsgi
+
+    async def request_validation(self, req, query, json, headers, cookies):
+        if query:
+            req.context.query = query(**req.params)
+        if headers:
+            req.context.headers = headers(**req.headers)
+        if cookies:
+            req.context.cookies = cookies(**req.cookies)
```
kemingy, line 232: To keep it consistent with other parts, it should be better to use `.parse_obj()`.

kemingy (commented):
Really appreciate your contribution! Thanks.

spectree/plugins/falcon_plugin.py
onecrayon: Good call! I caught that on one of the `json` calls, but missed it on these. I'll get that updated. I'm also having some trouble getting the async stuff working on a local testing project, so I'll let you know when I've got that sorted out.

onecrayon:
@kemingy This is probably ready for a look when you have time. I've got it working locally; not sure why the tests were working for me locally when it was actually completely broken, though.

kemingy (changes requested):
LGTM. Please help resolve the last comment.

The falcon 3 will check the media type. There is another PR #141 working on this one. That's the main breaking changes for falcon 3 plugins.

onecrayon:
Huh, I fixed the commented issue, but I must have forgotten to push it; I'll get that sorted out.

onecrayon:
@kemingy Lost track of this hard, but here's the requested change!

kemingy (approved):
Thanks for your help.

Issue #47, opened 2020-06-14
[BUG] Query parameters default values can be found in openapi.json but don't show up in docs.


zero-shubham:
Query parameters default values can be found in openapi.json but don't show up in docs.

![Docs](https://user-images.githubusercontent.com/17028205/84585708-1871dd00-ae30-11ea-80f1-1d7275474dbb.png)
![openapi.json](https://user-images.githubusercontent.com/17028205/84585709-1a3ba080-ae30-11ea-83fd-3ac6c1c50f57.png)

zero-shubham:
I have a fix, may i create a PR? 

Below is the needed code change, Idea is for query we will only pass on BaseModel, so it's properties should added in the schema rather than the BaseModel itself.
![Code to change](https://user-images.githubusercontent.com/17028205/84586182-e3b45480-ae34-11ea-8a35-47c39fca169d.png)

Below is the result.
![Selection_276](https://user-images.githubusercontent.com/17028205/84586200-15c5b680-ae35-11ea-90a3-dc88d45006bf.png)

kemingy:
Thanks for your feedback. I'm checking this problem.

This problem is more complex than I thought. We need to consider nested models. For example:

```py
class Language(str, Enum):
    en = 'en-US'
    zh = 'zh-CN'

class Header(BaseModel):
    Lang: Language
```

I feel that it still needs to be a schema instead of all the keys here.

zero-shubham:
Aha when query has to be an enum.. Got it. Will look into too. Thanks!

kemingy:
https://stackoverflow.com/a/46172865

This may help you. In swagger, default value is different from example values. While redoc can handle this very well.

zero-shubham:
In my PR i did a mistake, following is how "required" should be evaluated
![required_evaluation](https://user-images.githubusercontent.com/17028205/84589346-706a0d00-ae4b-11ea-81b2-18b90076747a.png)

By nested models if meant the following, I think it works fine. Sorry if I understood this wrong.
![Selection_277](https://user-images.githubusercontent.com/17028205/84589370-94c5e980-ae4b-11ea-85d6-ab9c4dbf8cb5.png)

![Selection_279](https://user-images.githubusercontent.com/17028205/84589371-9c858e00-ae4b-11ea-9807-de6f491f284d.png)

![Selection_280](https://user-images.githubusercontent.com/17028205/84589373-a1e2d880-ae4b-11ea-9f2c-f537d81ce019.png)

[...]

PR #48, opened 2020-06-14
fix query parsing

zero-shubham:
for issue [#47 ](https://github.com/0b01001001/spectree/issues/47)

### DIFF
--- a/spectree/utils.py
+++ b/spectree/utils.py
@@ -51,7 +51,7 @@ def parse_params(func, params, models):
             params.append({
                 'name': query_param,
                 'in': 'query',
-                'required': has_default,
+                'required': not has_default,
                 'schema': models[func.query]["properties"][query_param]
             })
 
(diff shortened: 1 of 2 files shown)


spectree/utils.py
```diff
@@ -44,14 +44,17 @@ def parse_params(func, params, models):
...
     if hasattr(func, 'query'):
-        params.append({
-            'name': func.query,
-            'in': 'query',
-            'required': True,
-            'schema': {
-                '$ref': f'#/components/schemas/{func.query}',
-            }
-        })
+        for query_param in models[func.query]["properties"].keys():
```
yedpodtrzitko, line 47: How about iterating over `.items()` instead of just `.keys()`? You can then replace `models[func.query]["properties"][query_param]` below just with the value from key/value pair you iterate over. Feels easier to read.

kemingy: check the current implementation. https://github.com/0b01001001/spectree/blob/49d85ec20b7c3756a774c10e32d7be248a7b2a14/spectree/utils.py#L49-L57

This PR is covered by #49.

I'll close it.

PR #52, opened 2020-07-01
Dev: add `before` and `after` hooks

### LINKED ISSUE #51: callback handler for the validation part
Sometimes people may need to do something else when there is a validation error, like recording the metric. Since the validation happens before the endpoint function, there should be a way to do this.

kemingy:
fix #51


spectree/spec.py
```diff
@@ -15,10 +17,22 @@ class SpecTree:
...
     :param kwargs: update default :class:`spectree.config.Config`
     """
 
-    def __init__(self, backend_name='base', backend=None, app=None, **kwargs):
+    def __init__(self,
+                 backend_name='base', backend=None,
+                 app=None,
+                 before=default_before_handler, after=default_after_handler,
+                 **kwargs):
+        self.before = before
```
anthonyfong100, line 34: Is it possible to move this to the validate function instead as an argument. This would allow for different callback function for each decorators.

kemingy: Good point, I'll support both.

PR #54, opened 2020-08-07
Add documentation-exclusive decorator

ruipgil:
Hi!

This is a great library, thank you!

I've used this lib quite a lot, but I prefer to validate the payloads manually.

With that in mind, I've created this PR which adds the `doc` decorator which does the same things as the `validate`, but without the validation part.

Let me know what you think, and if you'd consider merging it!

Thank you!

### DIFF
--- a/spectree/spec.py
+++ b/spectree/spec.py
@@ -66,11 +66,12 @@ class SpecTree:
                 return True
             return False
 
-    def validate(self, query=None, json=None, headers=None, cookies=None, resp=None, tags=[]):
+    def _base(
+        self, validator_func, query=None, json=None, headers=None, cookies=None, resp=None,
+        tags=None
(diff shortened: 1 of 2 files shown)


kemingy:
Thanks for your feedback.

Currently, this library is using `pydantic` which should be very fast (according to the [benchmark](https://pydantic-docs.helpmanual.io/benchmarks/)) and flexible. You should be able to add some customized validation code to the pydantic model with [validator](https://pydantic-docs.helpmanual.io/usage/validators/). Personally, I think this can make the code clean by separate the validation code and endpoint code.

Please let me know if you have any other concerns. It would be helpful if you can provide a scenario.

kemingy:
Feel free to reopen this if you have any other concerns.

PR #57, opened 2020-09-16
Add 422 default model

honeydev:
* Add default pydantic model for 422 error
* Remove current 422 error description


kemingy (commented):
Sorry for the late processing. Thanks for your contribution.

PR #70, opened 2020-10-23
Fixes for operationId, flask/get_json() and more unique path name

natasha-aleksandrova:
This PR includes fixes for following items:
* operationID -> operationId per openapi standard
* operationId value to be more unique (encountered clashes)
* fix for get_json() crash if application/json is set without json body
* also added `after` hook before calling `abort()` on validation error - we return standard json object on any error and needed to modify the default 422 error response in this lib.

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -143,6 +143,7 @@ class FlaskPlugin(BasePlugin):
 
         before(request, response, req_validation_error, None)
         if req_validation_error:
+            after(request, response, resp_validation_error, None)
             abort(response)
 
         response = make_response(func(*args, **kwargs))
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -111,18 +111,8 @@ class FlaskPlugin(BasePlugin):
         return ''.join(subs), parameters
 
     def request_validation(self, request, query, json, headers, cookies):
-        from werkzeug.exceptions import BadRequest
-
         req_query = request.args or {}
-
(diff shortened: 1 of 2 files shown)


spectree/plugins/flask_plugin.py
```diff
@@ -111,8 +111,18 @@ def parse_path(self, route):
...
-        req_json = request.get_json() or {}
+
+        try:
+            # if content-type is set to application/json but json is passed in below crashes
+            # it is technically intentional by flask, however this can have unexpected results for clients 
+            # that already set these properties incorrectly
+            req_json = request.get_json() or {}
+        except BadRequest:
+            req_json = {}
+
```
kemingy, line 125: Maybe can use `request.get_json(silent=True)`? So it will ignore the parsing errors and return `None`

spectree/plugins/flask_plugin.py
```diff
@@ -138,6 +148,7 @@ def validate(self,
 
         before(request, response, req_validation_error, None)
         if req_validation_error:
+            after(request, response, resp_validation_error, None)
```
kemingy, line 151: I guess it should be `after(request, response, req_validation_error, None)`

spectree/plugins/flask_plugin.py
natasha-aleksandrova: yes I like that better

spectree/plugins/flask_plugin.py
natasha-aleksandrova: yes you're right I think it should be `req_validation_error`

Issue #96, opened 2020-12-13
[Feature]Can't support list type swagger response


sunlin92:
like:
```python
class Item(BaseModel):
    name: str
    price: float


@app.route("/", method=["POST"])
@api.validate(json=Item, resp=Response(HTTP_200=List[Item]), tags=["demo"])
def demo():
    item = Item.parse_obj(request.context.json)
    return [item, item, item]
```

kemingy:
One possible way is to use `__root__` [pydantic __root__](https://pydantic-docs.helpmanual.io/usage/models/#custom-root-types).

```python
class Item(BaseModel):
    name: str
    price: float


class ItemList(BaseModel):
    __root__: List[Item]


@app.route("/", method=["POST"])
@api.validate(json=Item, resp=Response(HTTP_200=ItemList), tags=["demo"])
def demo():
    item = request.context.json
    return ItemList.parse_obj([item, item, item])
```

By the way, you don't need to parse `request.context.json` again. For the return statement, you can use `return ItemList.parse_obj([item, item, item])` or `return ItemList(__root__=[item, item, item])`.

sunlin92:
I really appreciate your help. 😄
Are there any plans to support this syntax（`HTTP_200=List[Item]`)？
It looks like be easier to read and use.

Thanks again

kemingy:
This syntax looks nice. But I feel it can be difficult to support. Because it means the library should be able to verify any types:

* `List[Item]`
* `Dict[str, Item]`
* `Dict[str, List[Item]]`
* ...

I'm not sure if there is an easy way to do this. Let me know if you have any ideas.

sunlin92:
FastAPI seems to support this syntax, I will create a pr when i complete it if needed.

kemingy:
Sure. Thanks for your help.

yoursvivek:
Data validation is easy. Pydantic provides a function to load data of any supported type or a combination thereof. 

One just needs to handle registry into `spec.models`.

https://pydantic-docs.helpmanual.io/usage/models/#parsing-data-into-a-specified-type

onecrayon:
Has anyone undertaken this issue yet? It’s something I need and I was wondering if I should take a swing at it or if it’s already underway.

kemingy:
> Has anyone undertaken this issue yet? It’s something I need and I was wondering if I should take a swing at it or if it’s already underway.

[...]

PR #100, opened 2021-01-07
enable view function annotation based type detection

yoursvivek:
The current implementation of spectree uses arguments to validate method for declaration of types for query params, request body, cookies and headers and in turn adds a context attribute to request object for all three framework implementations like following:

```python
# flask view implementation
@app.route("/api/user/<name>", methods=["POST"])
@api.validate(
    query=Query,
    json=JSON,
    cookies=Cookies,
    resp=Response(HTTP_200=Resp, HTTP_401=None),
    tags=["api", "test"],
    after=api_after_handler,
)
def user_score(name):
    score = [randint(0, request.context.json.limit) for _ in range(5)]
    score.sort(reverse=request.context.query.order)
    assert request.context.cookies.pub == "abcdefg"
    assert request.cookies["pub"] == "abcdefg"
    return jsonify(name=request.context.json.name, score=score)
```

While this is good, but static linters like mypy complain about `context` attribute missing from flask request object. If view functions itself are annotated with their type and we additionally inject kwargs in view, linters are happy and long `request.context.query.order` access is now just `query.order`. Following is an example of annotated view function:

```python
# flask view implementation with annotation
@app.route("/api/user_annotated/<name>", methods=["POST"])
@api.validate(
    resp=Response(HTTP_200=Resp, HTTP_401=None),
    tags=["api", "test"],
    after=api_after_handler,
)
def user_score_annotated(
    name,
    query: Query,
    json: JSON,
    cookies: Cookies
):
    score = [randint(0, json.limit) for _ in range(5)]
    score.sort(reverse=query.order)
    assert cookies.pub == "abcdefg"
    assert request.cookies["pub"] == "abcdefg"
    return jsonify(name=json.name, score=score)
```

Current implementation is an opt-in feature disabled by default. To opt-in simple instantiate `SpecTree` instance with an additional parameter like following:
```python
api = SpecTree("flask", annotations=True)
```

This feature doesn't change anything if not opted in. Once opted it, user can have a nice API.

I'm willing to discuss the merits, demerits of this approach. It's much closer to _FaskAPI_ usage and since we are already using `pydantic` and annotations inside `spectree`, why stop there.

All tests are in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ Check the [examples](/examples) folder.
 
 If the request doesn't pass the validation, it will return a 422 with JSON error message(ctx, loc, msg, type).
 
+### Opt-in type annotation feature
+This library also supports injection of validated fields into view function arguments along with parameter annotation based type declaration. This works well with linters that can take advantage of typing features like mypy. See examples section below.
+
 ## How To
 
 > How to add summary and description to endpoints?
@@ -65,7 +68,7 @@ Check the [pydantic](https://pydantic-docs.helpmanual.io/usage/schema/) document
 
 Of course. Check the [config](https://spectree.readthedocs.io/en/latest/config.html) document.
 
-You can update the config when init the spectree like: 
+You can update the config when init the spectree like:
 
 ```py
 SpecTree('flask', title='Demo API', version='v1.0', path='doc')
@@ -153,6 +156,25 @@ if __name__ == "__main__":
 
 ```
 
+#### Flask example with type annotation
+
+```python
+# opt in into annotations feature
+api = SpecTree("flask", annotations=True)
+
+
+@app.route('/api/user', methods=['POST'])
+@api.validate(resp=Response(HTTP_200=Message, HTTP_403=None), tags=['api'])
+def user_profile(json: Profile):
+    """
+    verify user profile (summary of this endpoint)
+
+    user's name, user's age, ... (long description)
+    """
+    print(json) # or `request.json`
+    return jsonify(text='it works')
+```
+
 ### Falcon
 
 ```py
@@ -199,6 +221,25 @@ if __name__ == "__main__":
     httpd = simple_server.make_server('localhost', 8000, app)
     httpd.serve_forever()
 
+```
+
+#### Falcon with type annotations
+
+```python
+# opt in into annotations feature
+api = SpecTree("flask", annotations=True)
+
+
+class UserProfile:
+    @api.validate(resp=Response(HTTP_200=Message, HTTP_403=None), tags=['api'])
+    def on_post(self, req, resp, json: Profile):
+        """
+        verify user profile (summary of this endpoint)
+
+        user's name, user's age, ... (long description)
+        """
+        print(req.context.json)  # or `req.media`
+        resp.media = {'text': 'it works'}
 ```
 
 ### Starlette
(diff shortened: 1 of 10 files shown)


yedpodtrzitko:
>  static linters like mypy complain about context attribute missing from flask request object

if this is done just to make `mypy` pass, how about acknowledge the `request` object isn't vanilla `flask.Request` and add custom `Request` class instead, something like this:


```diff 
# anywhere, eg. spectree/plugins/flask_plugin.py
+ from flask import Request, request as request_

+ class SpecTreeRequest(Request):
+     context: Context

+ request: SpecTreeRequest = request_
```

```diff
# myapp.py
- from flask import abort
+ from spectree.plugins.flask_plugin import request

# this will now pass without mypy error
payload = request.context.json
```

kemingy:
> if this is done just to make mypy pass, how about acknowledge the request object isn't vanilla flask.Request and add custom Request class instead, something like this:

This works well but it's limited to `flask`.

Personally, I think linter is very useful (although Python linter still has a long way to go). Meanwhile, **FastAPI** is quite popular. I think a lot of people will accept this annotation approach.

One possible problem is that users should avoid using these keywords as path variables.

One possible benefit is that it may make it easy to implement #96.

yoursvivek:
> if this is done just to make `mypy` pass, how about acknowledge the `request` object isn't vanilla `flask.Request` and add custom `Request` class instead, something like this:

This is not just about making `mypy` happy. I find annotation based approach much cleaner for developers. Once a developer has access to an object instance and it's type by virtue of function signature, he can focus on his code right away instead of trying to decipher which all decorators have injected some additional attributes somewhere in an otherwise common object.

> One possible problem is that users should avoid using these keywords as path variables.

During development, concern about having same keyword in path variables also came to my mind. I don't have a concrete answer to this. For now, if I were in that position, I'll pass them in params to `validate` in those select function where I can't rename path params for any particular reason (part of public api documentation etc.).

Another approach would be to provide option to alias params name for query/json/headers/cookies through `validate` method.

```python
app.route("/search/{query}/")
api.validate(..., alias={"query": "q", "json": "body"})
def viewfunc(query: str, q: QueryModel, body: BodyModel):
   ...
```

yedpodtrzitko:
> I find annotation based approach much cleaner for developers.

I'm not opposing that, I like the benefits type annotations provide too, but from your comment it seemed like the motivation for this is just to make mypy pass. 

> developer has access to an object instance and it's type by virtue of function signature

That's sure useful, but I'm not sure if the tradeoff of inflating each function with 4 parameters is worth it, considering some of them are barely used (how many views really use `cookies` or `headers`?) So how about passing there just the `Context` object (which encapsulate all the data structures as it is now), similarly as Django is passing `request` to its views? This could solve all the problems:

- no property injected into the request object (so mypy is happy)

- you can have almost direct access (with only 1 level of nesting) to all the data structures through a function's parameter

- function signature is not inflated by 4 parameters, so there's lower probability of name clashing

- it's more future-proof than having 4 parameters. If anything changes in the `Context`, you still have the same 1 parameter, and it can facade the changes internally (maybe it can be even lazily evaluated?). If there would be any change within the 4 parameters, like name change, or adding yet another parameter, you'd need to change all function signatures retroactively.

...in fact I can imagine passing `context` into the views could be the default (only) way how to do that, even at the price of backward incompatibility. Having this as a config option is sure non-intrusive, but it increases the complexity of the code  and _something something_ Zen of Python one way how to do things.

kemingy:
> During development, concern about having same keyword in path variables also came to my mind. I don't have a concrete answer to this. For now, if I were in that position, I'll pass them in params to validate in those select function where I can't rename path params for any particular reason (part of public api documentation etc.).

Thanks for your answer. You are right, users can still use the `validate` function if there are some conflicts. And I think that's enough.

yoursvivek:
> That's sure useful, but I'm not sure if the tradeoff of inflating each function with 4 parameters is worth it, considering some of them are barely used (how many views really use `cookies` or `headers`?) 

Actually the PR requires you to only put parameters for what you want, not all four.

So if you only want to parse `body` your view signature would be:
```python
@route('/keys')
@api.validate(resp=Response("HTTP_200"))
def viewfunc(json: BodyModel):
    return json.keys()
```

kemingy:
LGTM. Can you add some documents to the README?

yoursvivek:
Examples and a description added in README.

kemingy:
Thanks.

kemingy (approved):

PR #101, opened 2021-01-07
pre-commit hooks for git

yoursvivek:
Since we are already using LGTM lint checks for black and isort, we can use pre-commit and make them run automatically during commit.


kemingy:
Thanks for your contribution. I haven't tried `pre-commit` before. I'm not sure if this config is compatible with the current `Makefile`. Can you help me double-check this?

yoursvivek:
* black
* isort
* flake8 
* autoflake
are all have same settings as run from Makefile. Except `--in-place` for autoflake the rest were not replying on Makefile for configuration/settings.

kemingy:
Cool! Thanks.

PR #102, opened 2021-01-08
for flask use current_app (recommended way)

yoursvivek:
- For accessing app use current_app
- ~~For running test use app_context~~
- Enable tests to run outside `app_context` by allowing `api.spec` to be cached for flask app

Earlier spectree was attaching flask app instance to FlaskPlugin itself. While there is still the case of app attaching to spectree instance, but that would need a broader introspection as it's common api/code for flask, starlette and falcon. Flask uses threadlocal and proxy to request and app instance. I'm not very familiar with starlette and falcon so I'll not disturb them.

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -1,4 +1,3 @@
-from flask import current_app
 from pydantic import ValidationError
 
 from .base import BasePlugin, Context
@@ -10,6 +9,7 @@ class FlaskPlugin(BasePlugin):
     FORM_MIMETYPE = ("application/x-www-form-urlencoded", "multipart/form-data")
 
     def find_routes(self):
+        from flask import current_app
 
         if self.blueprint_state:
             excludes = [
@@ -39,6 +39,8 @@ class FlaskPlugin(BasePlugin):
         return False
 
     def parse_func(self, route):
+        from flask import current_app
+
         if self.blueprint_state:
             func = self.blueprint_state.app.view_functions[route.endpoint]
         else:
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -1,3 +1,4 @@
+from flask import current_app
 from pydantic import ValidationError
 
 from .base import BasePlugin, Context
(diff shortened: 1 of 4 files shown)


spectree/plugins/flask_plugin.py
```diff
@@ -1,3 +1,4 @@
+from flask import current_app
```
kemingy, line 1: With this import, users that are using `falcon` or `starlette` will need to install `flask` as a dependency.

kemingy: According to [PEP8 imports](https://www.python.org/dev/peps/pep-0008/#imports):

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

I agree that this should be the recommended way.

So I'm thinking maybe we should let the users import plugins they need instead of the library imports everything.

The interface may look like this:

```py
api = SpecTree(FlaskPlugin)
```

This should be more intuitive than a string.

yoursvivek: I see that now. It didn't cross my mind why imports were withing the function. I'll fix that right away.

And I definitely like the later approach where uses passes plugin class to SpecTree init.

tests/test_plugin.py
```diff
@@ -3,11 +3,12 @@
 from .common import JSON, Cookies, Headers, Query, Resp, get_paths
 from .test_plugin_falcon import api as falcon_api
 from .test_plugin_flask import api as flask_api
+from .test_plugin_flask import flask_ctx  # noqa: F401
 from .test_plugin_starlette import api as starlette_api
 
 
 @pytest.mark.parametrize("api", [flask_api, falcon_api, starlette_api])
-def test_plugin_spec(api):
+def test_plugin_spec(api, flask_ctx):  # noqa: F811
```
kemingy, line 11: No need to add a new `flask_ctx` here. The `spec` property will be cached if you have called it once before.

This can be solved by adding the following code into `tests/test_plugin_flask.py`:

```py
with app.app_context():
    api.spec
```

yoursvivek: Implemented as suggested.

kemingy:
LGTM. Thanks for your contributions. I think `falcon` and `starlette` don't have this problem.

PR #105, opened 2021-01-27
Update readme with additional instruction about falcon respnse validation

FerdinandZhong:
An additional instruction about falcon response. 

Falcon response has both media and body(text in falcon 3.0). Spectree is only validating the media as it's meaningless to validate against string. Therefore I updated the Readme with the additional instruction to make this part more clear to users. And also fix a typo in Readme. :)


kemingy (commented):
LGTM

PR #106, opened 2021-02-11
add explicit_name parameter to explicitly name an API endpoint

arthurlenoir:


kemingy:
Hi, thanks for your contribution.

I guess the feature you want is to overwrite the endpoint name. Actually, this can be done through the `summary`, which is retrieved from the function/class comments.

For example:

```py
@spectree.validate()
def index():
    """endpoint alias"""
    pass
```

Then in the generated OpenAPI document, it will display "endpoint alias" instead of "index <GET>".

PR #110, opened 2021-03-16
Fix component to avoid loss of models when multiple spec-files

Vlczech:


kemingy:
There are some historical reasons. So `pydantic` still generates schemas with `definition` when there is a nested model.

For example:

```python
from typing import List


class Item(BaseModel):
    name: str


class ItemList(BaseModel):
    __root__: List[Item]


print(Item.schema())
print(ItemList.schema())
```

As we know, OpenAPI 3 doesn't have this property. So I just extract the definitions from the schemas and add them to `"#/components/schemas/`.

Let's get back to this PR. I think you are on the right way. There are 2 ways to fix it:

1. remove `del schema["definitions"]`, redundant definitions doesn't cause any error (need to fix the test cases)
2. deep copy instead of manipulating the original `self.models` dictionary
  * only copy the values
  * extract the definitions and add them to the schemas

So, what do think?

kemingy:
Close this one since it's fix by #133

PR #127, opened 2021-05-03
Add Authentication and Authorization logic.

AndreiDrang:


kurkl (approved):

spectree/models.py
```diff
@@ -44,3 +45,19 @@ class UnprocessableEntity(BaseModel):
...
+
+
+class SecureType(str, Enum):
+    HTTP = "http"
+    API_KEY = "apiKey"
+    OAUTH_TWO = "oauth2"
+    OPEN_ID_CONNECT = "openIdConnect"
+
+
+class SecuritySchemesData(BaseModel):
```
kemingy, line 57: Can export this to `spectree/__init__.py` file for users to import.

spectree/config.py
```diff
@@ -63,3 +68,6 @@ def update(self, **kwargs):
 
         assert self.UI in self._SUPPORT_UI, "unsupported UI"
         assert self.MODE in self._SUPPORT_MODE, "unsupported MODE"
+        if self.AUTH_METHODS:
+            for value in self.AUTH_METHODS.values():
+                SecuritySchemesData(**value)
```
kemingy, line 73: Will it be better to accept the instance of `SecuritySchemesData`? So when the user creates this, the data is already validated. Then we can add another method to get the JSON schema for OpenAPI to use.

kemingy (commented):
Awesome! Really appreciate your contribution.

Another thing I want to talk about is, for Authentication Error, should we return ValidationError(422) or AuthenticationError(401 or 403)?

spectree/config.py
AndreiDrang: Also, I'm not sure about `self.AUTH_METHODS` naming, maybe name from the OpenAPI scheme will be better?
For example - `self.securitySchemes`, not a pythonic style but... it is easy to understand what the parameter it is.

kemingy: Agree. `self.SECURITY_SCHEMES` sounds better. BTW, I'm thinking if we should use `pydantic.BaseSettings` for the config class.

AndreiDrang: > pydantic.BaseSettings

Sounds good but for some new PR.

AndreiDrang:
> Another thing I want to talk about is, for Authentication Error, should we return ValidationError(422) or AuthenticationError(401 or 403)?

I'm not sure that hardcoded Errors are good, maybe some flexibility is better.

Cause, for example in my case I not use 422 error from swagger config, I always return 200 code and put error info in payload under specific key, and frontend developers ask me questions about "422 error from swagger" and when it rises :smiley:

But I think, that `401` error will be better. Or we can just print some `WARNING` message(like SQLAlchemy) if user sets `security` key for route, but not set `401` or `403` error in resp.
image

spectree/models.py
```diff
@@ -1,7 +1,11 @@
+import re
 from enum import Enum
 from typing import Any, Dict, Sequence
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, validator
+
+# OpenAPI names validation regexp
+OAPI_NAME_RE = re.compile(r"^[A-Za-z0-9-._]+")
```
kemingy, line 8: Can choose one from (`OpenAPI_NAME_RE`, `OPENAPI_NAME_RE`). `OAPI_NAME_RE` is a little bit confusing.

spectree/models.py
```diff
@@ -54,10 +58,38 @@ class SecureType(str, Enum):
...
+
+class SecurityScheme(BaseModel):
+    """
+    Named security scheme
+    """
+
+    name: str = Field(
+        ...,
+        description="Custom security scheme name. Can only contain - [A-Za-z0-9-._]",
+    )
```
kemingy, line 85: What's the difference between `SecurityScheme.name` and `SecuritySchemeData.name`?

spectree/models.py
```diff
@@ -54,10 +58,38 @@ class SecureType(str, Enum):
...
+    """
+    Security scheme data
+    """
+
     type: SecureType = Field(..., description="Secure scheme type")
     name: str = None
     field_in: str = Field(None, alias="in")
     scheme: str = None
     openIdConnectUrl: str = None
     flows: dict = None
```
kemingy: Except `type`, all the other items are optional. So if the user create `SecuritySchemeData(type="http")`, should that be accepted?

kemingy:
> I'm not sure that hardcoded Errors are good, maybe some flexibility is better.
> 
> Cause, for example in my case I not use 422 error from swagger config, I always return 200 code and put error info in payload under specific key, and frontend developers ask me questions about "422 error from swagger" and when it rises 😃
> 
> But I think, that `401` error will be better. Or we can just print some `WARNING` message(like SQLAlchemy) if user sets `security` key for route, but not set `401` or `403` error in resp.

We can use `401` as the default one. I also provide `before_hook` for the user to change the request validation error response body.

spectree/models.py
AndreiDrang: > SecurityScheme.name

Custom name for auth scheme, init in `components->securitySchemes` and then can be used in `route->method->security`. Need validation.

> SecuritySchemeData.name

Custom variable name for header, for example. Validation not needed.

Screenshot_2021-05-05_21-39-58

AndreiDrang: Add more descriptions

spectree/models.py
AndreiDrang: Fix

kemingy:
Please run `make lint` and `make test` before the push.

Qparrot:
Hello, 
Do you know if it is now possible to add Authorization logic ? 
When I search for "security_scheme", "authorization" I found no related page on the [SpecTree documentation](https://spectree.readthedocs.io/en/latest/).

thank you for all the work :), Spectree is great!

kemingy:
> Do you know if it is now possible to add Authorization logic ?

@Qparrot I haven't added docs for this part. But you can refer to the test file https://github.com/0b01001001/spectree/blob/f5e56e57d580cdc78d67af20737c0a346246d595/tests/test_plugin_flask.py#L38 and the examples on the bottom of that file.

kemingy:
> Hello,
> Do you know if it is now possible to add Authorization logic ?
> When I search for "security_scheme", "authorization" I found no related page on the [SpecTree documentation](https://spectree.readthedocs.io/en/latest/).

The new PR #151 added the related document in the README. @Qparrot

PR #140, opened 2021-06-02
Fix securityschemes

Vlczech:
Fixes following error:

```
Structural error at components.securitySchemes
should be object
```


kemingy (commented):
Thanks

PR #141, opened 2021-06-02
Fix import statement compatibility in Falcon2 and 3

mortymacs:
As Falcon3 supports both `HTTPUnsupportedMediaType` and `MediaNotFoundError` exceptions, we need to change the order of the import statements to import Falcon3 then Falcon2.


mortymacs:
This is actually a hotfix of the last changes.

kemingy:
Can you provide some examples that will trigger the following exceptions:

* falcon 2 `MediaNotFoundError `
* falcon 3 `HTTPUnsupportedMediaType `
* falcon 3 `MediaNotFoundError `

Previously, I added `HTTPUnsupportedMediaType` to fix the falcon 3 compatible problems in #116. I'm not sure if falcon 2 need to deal with this kind of exceptions.

mortymacs:
> Can you provide some examples that will trigger the following exceptions:
> 
> * falcon 2 `MediaNotFoundError `
> * falcon 3 `HTTPUnsupportedMediaType `
> * falcon 3 `MediaNotFoundError `
> 
> Previously, I added `HTTPUnsupportedMediaType` to fix the falcon 3 compatible problems in #116. I'm not sure if falcon 2 need to deal with this kind of exceptions.

Sure, I think I can even update tests for it but I have to check the possibility. About Falcon2, that's why I added an `if/else` for the `MediaNotFoundError` exception.

PR #145, opened 2021-06-03
FIX: empty path enpoints leads to Semantic error => clean empty paths

Vlczech:
* If multiple OAS files are created, paths defined in other scope are empty in current OAS file. If it is an endpoint with path parameter, Semantic error occurs.
* Example: 
```
Semantic error at paths ./v1/foo/bar/{foo_bar_id}
Declared path parameter foo_bar_id needs to be defined as a path parameter at either the path or operation level
```


kemingy (commented):
Thanks

PR #146, opened 2021-06-03
FIX: Code structure (paths) leaking => short hash used instead

Vlczech:
(short hashes are used as they are used in GIT commits)


kemingy (commented):
Thanks

PR #149, opened 2021-06-15
Global security and some changes in security

Vlczech:
Possibility to define security at global (root) level or local (operational) level. It is also possible to override global (root) level by local (operational) level.

> [..] After you have defined the security schemes in the securitySchemes section, you can apply them to the whole API or individual operations by adding the security section on the root level or operation level, respectively. When used on the root level, security applies the specified security schemes globally to all API operations, unless overridden on the operation level. [..]

[Step 2. from [https://swagger.io/docs/specification/authentication/](https://swagger.io/docs/specification/authentication/)]


kemingy (approved):
Thanks.

PR #154, opened 2021-06-23
Servers definition

Vlczech:
Allows to define servers with url, description and variables as is defined here: https://swagger.io/docs/specification/api-host-and-base-path/

### DIFF
--- a/spectree/spec.py
+++ b/spectree/spec.py
@@ -264,6 +264,11 @@ class SpecTree:
             },
         }
 
+        if self.config.SERVERS:
+            spec["servers"] = [
+                server.dict() for server in self.config.SERVERS
+            ]
+
         if self.config.SECURITY:
             spec["security"] = [
                 {security_name: security_config}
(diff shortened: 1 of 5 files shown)


spectree/spec.py
```diff
@@ -264,6 +264,16 @@ def _generate_spec(self):
...
 
+        if self.config.SERVERS:
+            spec["servers"] = [
+                {
+                    "url": server.url,
+                    "description": server.description,
+                    "variables": server.variables,
+                }
+                for server in self.config.SERVERS
+            ]
```
kemingy, line 275: Is it better to use `[server.dict() for server in self.config.SERVERS]`?

kemingy (commented):
Thanks for your contribution.

kemingy (approved):
LGTM

spectree/spec.py
Vlczech: It seems good - I used vars or something like this previously, but the test failed due to different key positions. But this is OK :-)

PR #168, opened 2021-09-22
Fixed structural error in OAS

Vlczech:
If some item of server is not filled, structural error in OAS occured -> only filled items are generated to OAS specification.

obrazek

### DIFF
--- a/spectree/spec.py
+++ b/spectree/spec.py
@@ -255,7 +255,9 @@ class SpecTree:
         }
 
         if self.config.SERVERS:
-            spec["servers"] = [server.dict() for server in self.config.SERVERS]
+            spec["servers"] = [
+                server.dict(exclude_none=True) for server in self.config.SERVERS
+            ]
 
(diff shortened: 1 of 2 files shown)


spectree/spec.py
```diff
@@ -255,7 +255,11 @@ def _generate_spec(self):
         }
 
         if self.config.SERVERS:
-            spec["servers"] = [server.dict() for server in self.config.SERVERS]
+            spec["servers"] = []
+            for server in self.config.SERVERS:
+                spec["servers"].append(
+                    {k: v for k, v in server.dict().items() if v is not None}
+                )
```
kemingy, line 262:
```suggestion
            spec["servers"] = [
                server.dict(exclude_none=True) for server in self.config.SERVERS
            ]
```

kemingy (approved):

spectree/spec.py
Vlczech: OK, adjusted as you wish :)

PR #75, opened 2020-10-25
Customize UI from principal project Falcon 

oscfrayle:
# clean url:

 /api/v1/docs

# UI Config, the UI can be in the principal project

api = SpecTree(
    'falcon',
    title=TITLE,
    version=VERSION,
    path=PATH,
    ui=PAGES["swagger"]  # redoc
)

# UI Config, without param " ui ", native library default 
api = SpecTree(
    'falcon',
    title=TITLE,
    version=VERSION,
    path=PATH
)

PR #180, opened 2021-11-04
Add option for using custom page templates

robertmrk:
This PR adds a `PAGE_TEMPLATE` configuration option, that allows the users of the library to specify custom documentation page templates. If nothing is specified, the default `swagger` and `redoc` templates are used.

The ability to use custom page templates comes in handy if there is a need to customize Swagger UI for example.

I'm aware of the fact that #75 already exists, and it probably tries to achieve the same goal as this PR. But, the big difference between the two solutions is that the changes in this PR, allow the usage of multiple documentation pages, while the solution in #75 only allows a single documentation page when a custom value is specified. Additionally, in #75 the changes are only implemented for the Falcon plugin.

### DIFF
--- a/spectree/config.py
+++ b/spectree/config.py
@@ -2,6 +2,7 @@ import logging
 from typing import List, Optional
 
 from .models import SecurityScheme, Server
+from .page import DEFAULT_PAGE_TEMPLATES
 
 
 class Config:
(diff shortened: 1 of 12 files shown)


kemingy (approved):
LGTM. Thanks for your contribution!

Issue #131, opened 2021-05-14
Openapi description for path variable


parisni:
So far, I can't find a way to add a description for `path` variables. On the flask side, It would be cool to have a parameter such `doc` as [flask-restfull](https://flask-restplus.readthedocs.io/en/stable/swagger.html#documenting-with-the-api-doc-decorator) provides : 
```
@api.route('/my-resource/<id>', endpoint='my-resource')
@api.doc(params={'id': 'An ID'})
```

kemingy:
Check this example: https://github.com/0b01001001/spectree/blob/36c4d18834eb2bbe19098d97e24036baca3943e2/examples/flask_demo.py#L55-L76

For the path variables, this library doesn't provide a strong validation with `pydantic`. It will use the default path converter.

But of course you can still get the correct OpenAPI document for path variables.

parisni:
> But of course you can still get the correct OpenAPI document for path variables

Sorry I cannot find out in your example how to add a description to `source`  and `target` path variable. I can see how to add constraints on them such as length but no description field that would appear in the openapi specs

kemingy:
Sorry I misunderstood your question before. Currently, this library doesn't support descriptions for the path variables. You can use function description for the path variables.

One possible way is to add another path model. But I'm not sure if we should validate the input again.

```py
class PathVariables(pydantic.BaseModel):
    source: str = pydantic.Field(..., description='source language')
    target: str

@api.route('/api/<string:source>/<string:target>')
@spec.validate(path=PathVariables)
def index(source, target):
    pass
```

Feel free to give more feedbacks about this.

PR #181, opened 2021-11-05
Add option to describe path parameters

robertmrk:
This PR adds a new `path_parameter_descriptions` parameter to `SpecTree.validate`. This parameter can be used to provide a description of the endpoint's path parameters. The key is interpreted as the name of the path variable, while the value is used as the description.

Example endpoint:
```python
@app.route('/foo/<foo>/bar/<bar>', methods=['GET'])
@api.validate(path_parameter_descriptions={"foo": "Foo's description."})
def foo(foo, bar):
    pass
```
The generated spec:
```json
"/foo/{foo}/bar/{bar}": {
  "get": {
    "summary": "foo <GET>",
    "operationId": "get_/foo/{foo}/bar/{bar}",
    "description": "",
    "tags": [
      
    ],
    "parameters": [
      {
        "name": "foo",
        "in": "path",
        "required": true,
        "schema": {
          "type": "string"
        },
        "description": "Foo's description."
      },
      {
        "name": "bar",
        "in": "path",
        "required": true,
        "schema": {
          "type": "string"
        },
        "description": ""
      }
    ],
    "responses": {
      
    }
  }
}
```

This PR is also a solution for #131. But instead of providing the description of path variables using a pedantic model, it uses just a plain dictionary, since path parameters are already validated by web frameworks.

### DIFF
--- a/spectree/plugins/base.py
+++ b/spectree/plugins/base.py
@@ -48,9 +48,11 @@ class BasePlugin:
         """
         raise NotImplementedError
 
-    def parse_path(self, route):
+    def parse_path(self, route, path_parameter_descriptions):
         """
         :param route: API routes
+        :param path_parameter_descriptions: A dictionary of path parameter names and
+            their description.
 
         parse URI path to get the variables in path
         """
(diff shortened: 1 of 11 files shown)


kemingy (approved):
Thanks.

PR #182, opened 2021-11-07
Add option to specify custom descriptions for response status codes.

robertmrk:
Hello folks!

The default descriptions provided by the library for the HTTP response status codes are usually good enough to understand the reason for a given response. However, there are some status codes whose meaning might not be so clear. A good example for this might be `409`, which indicates that the request could not be processed because of conflict in the current state of the resource. However, the possible reasons for a conflict might be different for different endpoints. For this reason, it would be very useful to have the option to specify custom descriptions for response status codes.

This PR adds the option to specify response models on endpoints together with a custom description.

```python
@app.route('/foo', methods=['GET'])
@api.validate(
    resp=Response(
        HTTP_200=Message,
        HTTP_201=(Message, "Custom description."),
    )
)
def foo():
    pass
```
The example above results in the following OpenAPI spec output:
```json
"/foo": {
  "get": {
    "summary": "foo <GET>",
    "operationId": "get_/foo",
    "description": "",
    "tags": [
      
    ],
    "parameters": [
      
    ],
    "responses": {
      "200": {
        "description": "OK",
        "content": {
          "application/json": {
            "schema": {
              "$ref": "#/components/schemas/b28b7af.Message"
            }
          }
        }
      },
      "201": {
        "description": "Custom description.",
        "content": {
          "application/json": {
            "schema": {
              "$ref": "#/components/schemas/b28b7af.Message"
            }
          }
        }
      },
      "422": {
        "description": "Unprocessable Entity",
        "content": {
          "application/json": {
            "schema": {
              "$ref": "#/components/schemas/6a07bef.ValidationError"
            }
          }
        }
      }
    }
  }
},

```

### DIFF
--- a/spectree/response.py
+++ b/spectree/response.py
@@ -1,33 +1,58 @@
-from typing import Type
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Type, Union
 
 from pydantic import BaseModel
 
 from .utils import get_model_key, parse_code
+
+OptionalModelType = Optional[Type[BaseModel]]
 
 
 class Response:
     """
     response object
 
-    :param codes: list of HTTP status code, format('HTTP_[0-9]{3}'), 'HTTP200'
-    :param code_models: dict of <HTTP status code>: <`pydantic.BaseModel`> or None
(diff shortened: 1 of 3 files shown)


kemingy (approved):
Thanks

PR #185, opened 2021-12-13
Form data support

Michae1Weiss:
Hi everyone!

This PR shows an example of support for uploading files via Swagger.

Usage example with ```werkzeug.datastructures.FileStorage``` (Flask):

```python3
from werkzeug.datastructures import FileStorage
from spectree import models

class File(BaseModel):
    uid: str
    file: models.BaseFile

@api.validate(form_data=File)
def with_file():
    file_dict = request.context.form_data.file.dict()
    file = FileStorage(**file_dict)
    return jsonify()
...
```

## Issues:
1. If uploading files using the Falcon framework plugin - ```content_length``` is always 0.

## Questions:
1. How to get ```content_length``` from ```falcon.cyutil.reader.BufferedReader``` object?
2. Initially ```SpecTree.validate``` only accepts ```json``` parameter to define request body. Would it be correct to separate request body schemas for ```multipart/form-data``` and ```application/json``` content types with different parameters (```json``` and ```from_data```) like I did or is there a cleaner way to do it?</br>
For example, the 'drf-yasg' module uses a generic ```request_body``` parameter for both content types.
To define the content type **'drf-yasg'** parses the view class object and if a view include only form parsers - swagger will support file upload.
[**drf-yasg implementation**](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/utils.py#L366)</br>
Wouldn't it make sense to make a generic ```request_body``` parameter like in **'drf-yasg**' and understand expected request content type of the request body by some flag for example?

### DIFF
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -188,16 +188,17 @@ class FalconPlugin(BasePlugin):
                 # Compare first part of content_type with media types
                 if part.content_type.split('/')[0] == 'text':
                     data[part.name] = part.text
-                elif part.content_type.split('/')[0] in ['image', 'application']:
+                elif part.content_type.split('/')[0] in ['image', 'application', 'audio']:
                     data[part.name] = {
                         "filename": part.filename,
                         "name": part.name,
-                        "content_length": len(part.data),
+                        # "content_length": len(part.data),  # FIXME: raises error if file is too big
+                        "content_length": 0,  # TODO: replace placeholder
                         "mimetype": part.content_type,
-                        "stream": part.data
+                        "stream": part.stream.read()
                     }
                 # TODO: add support for other media types?
-                # TODO: handle unsupported media types?
+                # TODO: or handle unsupported media types?
             req.context.form_data = form_data.parse_obj(data)
 
     def validate(
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -167,7 +167,7 @@ class FalconPlugin(BasePlugin):
 
         return f'/{"/".join(subs)}', parameters
 
-    def request_validation(self, req, query, json, headers, cookies):
+    def request_validation(self, req, query, json, form_data, headers, cookies):
         if query:
             req.context.query = query.parse_obj(req.params)
         if headers:
@@ -188,6 +188,7 @@ class FalconPlugin(BasePlugin):
         func,
         query,
         json,
+        form_data,
         headers,
         cookies,
         resp,
@@ -201,7 +202,7 @@ class FalconPlugin(BasePlugin):
         _self, _req, _resp = args[:3]
         req_validation_error, resp_validation_error = None, None
         try:
-            self.request_validation(_req, query, json, headers, cookies)
+            self.request_validation(_req, query, json, form_data, headers, cookies)
             if self.config.ANNOTATIONS:
                 for name in ("query", "json", "headers", "cookies"):
                     if func.__annotations__.get(name):
@@ -242,7 +243,7 @@ class FalconAsgiPlugin(FalconPlugin):
     OPEN_API_ROUTE_CLASS = OpenAPIAsgi
     DOC_PAGE_ROUTE_CLASS = DocPageAsgi
 
-    async def request_validation(self, req, query, json, headers, cookies):
+    async def request_validation(self, req, query, json, form_data, headers, cookies):
         if query:
             req.context.query = query.parse_obj(req.params)
         if headers:
@@ -263,6 +264,7 @@ class FalconAsgiPlugin(FalconPlugin):
         func,
         query,
         json,
+        form_data,
         headers,
         cookies,
         resp,
@@ -276,7 +278,7 @@ class FalconAsgiPlugin(FalconPlugin):
         _self, _req, _resp = args[:3]
         req_validation_error, resp_validation_error = None, None
         try:
-            await self.request_validation(_req, query, json, headers, cookies)
+            await self.request_validation(_req, query, json, form_data, headers, cookies)
             if self.config.ANNOTATIONS:
                 for name in ("query", "json", "headers", "cookies"):
                     if func.__annotations__.get(name):
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -182,6 +182,23 @@ class FalconPlugin(BasePlugin):
             media = None
         if json:
             req.context.json = json.parse_obj(media)
+        elif form_data:
+            data = {}
+            for part in media:
+                # Compare first part of content_type with media types
+                if part.content_type.split('/')[0] == 'text':
+                    data[part.name] = part.text
+                elif part.content_type.split('/')[0] in ['image', 'application']:
+                    data[part.name] = {
+                        "filename": part.filename,
+                        "name": part.name,
+                        "content_length": len(part.data),
+                        "mimetype": part.content_type,
+                        "stream": part.data
+                    }
+                # TODO: add support for other media types?
+                # TODO: handle unsupported media types?
+            req.context.form_data = form_data.parse_obj(data)
 
     def validate(
         self,
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -182,9 +182,9 @@ class FalconPlugin(BasePlugin):
             media = None
         if json:
             req.context.json = json.parse_obj(media)
-        elif form_data:
+        elif form_data and media is not None:
             data = {}
-            for part in media:
+            for part in media:  # TODO: add support for falcon 2.0 (media is always None for 'multipart/form-data')
                 # Compare first part of content_type with media types
                 if part.content_type.split('/')[0] == 'text':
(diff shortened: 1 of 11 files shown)


spectree/plugins/falcon_plugin.py
```diff
@@ -182,12 +182,32 @@ def request_validation(self, req, query, json, headers, cookies):
...
+                        "filename": part.filename,
+                        "name": part.name,
+                        # "content_length": len(part.data),  # FIXME: raises error if file is too big
+                        # sadly, BufferedReader don't have 'tell' method
+                        "content_length": 0,  # TODO: replace placeholder with something that will work
+                        "content_type": part.content_type,
+                        "stream": part.stream.read(),
+                    }
+                # TODO: add support for other media types?
+                # TODO: OR handle unsupported media types?
```
kemingy, line 202: Refs: https://www.iana.org/assignments/media-types/media-types.xhtml

How about the following?

* font
* video
* message
* model

I wonder what's the major difference among these types for file handlers?

spectree/plugins/falcon_plugin.py
```diff
@@ -182,12 +182,32 @@ def request_validation(self, req, query, json, headers, cookies):
...
+                    data[part.name] = part.text
+                elif part.content_type.split('/')[0] in ['image', 'application', 'audio']:
+                    data[part.name] = {
+                        "filename": part.filename,
+                        "name": part.name,
+                        # "content_length": len(part.data),  # FIXME: raises error if file is too big
+                        # sadly, BufferedReader don't have 'tell' method
+                        "content_length": 0,  # TODO: replace placeholder with something that will work
+                        "content_type": part.content_type,
+                        "stream": part.stream.read(),
```
kemingy, line 199: Since we already consume the stream buffer, why not just use `part.data` to access the data and content length?

I'm not sure if it's possible to keep the buffer and let the user consume it later.

spectree/plugins/falcon_plugin.py
```diff
@@ -182,12 +182,32 @@ def request_validation(self, req, query, json, headers, cookies):
             media = None
         if json:
             req.context.json = json.parse_obj(media)
+        elif form_data and media is not None:
+            data = {}
+            for part in media:  # TODO: add support for falcon 2.0 (media is always None for 'multipart/form-data')
```
kemingy, line 187: BTW, async falcon should also be in the TODO list.

spectree/plugins/starlette_plugin.py
```diff
@@ -37,10 +38,28 @@ def register_route(self, app):
...
+                        "stream": value.file.read(),  # FIXME: takes a long time if the file is large
+                        "content_length": value.file.tell(),
+                    }
+                else:
+                    data[key] = value
+
         request.context = Context(
             query.parse_obj(request.query_params) if query else None,
             json.parse_raw(await request.body() or "{}") if json else None,
+            form_data.parse_obj(data or "{}") if form_data and data else None,
```
kemingy, line 62:
```suggestion
            form_data.parse_obj(data or {}) if form_data and data else None,
```

kemingy:
Thanks for your great work!

You can fix the lint by running `make format` and test with `make lint`.

Michae1Weiss:
@kemingy Thanks for review! I’ve added form-data support for async falcon and implemented various optimizations and fixes based on your comments. It's now time to write some tests for it :)
I will try to write them next week.
I'd be grateful for a second review and feedback. I'm not quite sure about some of the fixes (still need to learn more about asgi and cgi)

kemingy:
Hi, @Michae1Weiss. Thanks for your contribution. It may take me some time to review the code. I will try to give you some feedback by the end of this week.

spectree/plugins/starlette_plugin.py
```diff
@@ -38,9 +39,27 @@ def register_route(self, app):
             )
 
     async def request_validation(self, request, query, json, form_data, headers, cookies):
+        data = {}
+        if request.headers.get('content-type', '').split(";")[0] in self.FORM_MIMETYPE:
```
yihong0618: if get the default `''` and split here index[0]  will raise an error

spectree/plugins/flask_plugin.py
```diff
@@ -122,30 +122,34 @@ def parse_path(self, route, path_parameter_descriptions):
...
+                    if v.get('type') == 'array':
+                        value = req_form.getlist(k)
+                    else:
+                        value = req_form.get(k)
+                    if value is not None:
+                        form_dict[k] = value
+                form_dict.update(**request.files.to_dict())
+                if form:
+                    req_form = form_dict
+                elif json:
```
yihong0618, line 146: maybe if here?

spectree/plugins/starlette_plugin.py
Michae1Weiss: Thanks for the feedback!
You're right, splitting an empty string with a `None` separator returns [] and `''.split()[0]` will raise an error.
But in this case `split()` uses a separator `;`. So `''.split(';')[0]` will return an empty string
From Python documentation:
`Splitting an empty string with a specified separator returns [''].`
Refs: https://docs.python.org/3/library/stdtypes.html#str.split

yihong0618: got your point, you are right, cool~

spectree/plugins/falcon_plugin.py
```diff
@@ -167,27 +170,59 @@ def parse_path(self, route, path_parameter_descriptions):
...
+                media = None
             req.context.json = json.parse_obj(media)
+        elif form:
+            # This must be done to avoid a bug in cgi.FieldStorage.
+            # https://gist.github.com/kgriffs/fe371bfbb2d63211889b
+            req.env.setdefault("QUERY_STRING", "")
+
+            try:
+                _form = cgi.FieldStorage(fp=req.stream, environ=req.env)  # TODO: rename
+            except ValueError:  # Invalid boundary?
```
kemingy, line 211: Will [`falcon.Request.bounded_stream`](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html?highlight=file#falcon.Request.bounded_stream) be better to avoid this exception?

spectree/plugins/falcon_plugin.py
```diff
@@ -167,27 +170,59 @@ def parse_path(self, route, path_parameter_descriptions):
...
+        if isinstance(field, list):
+            return [self.parse_field(subfield) for subfield in field]
+
+        # When file name isn't ascii FieldStorage will not consider it.
+        encoded = field.disposition_options.get("filename*")
+        if encoded:
+            encoding, filename = encoded.split("''")
+            field.filename = filename
+            field.file = BytesIO(field.file.read().encode(encoding))
+        if getattr(field, "filename", False):
```
kemingy, line 183: I think `getattr(field, "filename", False)` will always return `None` since the type of `field` is always `cgi.FieldStorage`.

spectree/plugins/falcon_plugin.py
```diff
@@ -167,27 +170,59 @@ def parse_path(self, route, path_parameter_descriptions):
 
         return f'/{"/".join(subs)}', parameters
 
-    def request_validation(self, req, query, json, headers, cookies):
+    def parse_field(self, field):
+        if isinstance(field, list):
+            return [self.parse_field(subfield) for subfield in field]
+
+        # When file name isn't ascii FieldStorage will not consider it.
+        encoded = field.disposition_options.get("filename*")
```
kemingy, line 178: Is `filename*` a regex?

spectree/plugins/falcon_plugin.py
```diff
@@ -167,27 +170,59 @@ def parse_path(self, route, path_parameter_descriptions):
...
+        if encoded:
+            encoding, filename = encoded.split("''")
```
kemingy, line 180: Any example for this one? I'm not sure when will the `field.filename` contains `''`.

kemingy (commented):
Thanks for your great work. Feel free to leave some comments. Especially for the `cgi.FileStorage` part. I'm not very familiar with it. I'm not sure if I understand your code correctly.

spectree/plugins/flask_plugin.py
```diff
@@ -122,21 +122,34 @@ def parse_path(self, route, path_parameter_descriptions):
...
-                )
-        else:
-            req_json = request.get_json(silent=True) or {}
+        req_json = request.get_json(silent=True) or {}
+        req_form = request.form or {}  # or MultiDict() ?
         req_headers = request.headers or {}
         req_cookies = request.cookies or {}
+
+        if request.mimetype in self.FORM_MIMETYPE:
+            body = form or json or None
```
yedpodtrzitko, line 133: what is `or None` for?

spectree/plugins/flask_plugin.py
yedpodtrzitko: elsewhere in this PR there is a condition that will raise exception if both `json` and `form` exist, so it seems like they are mutually exclusive

spectree/plugins/falcon_plugin.py
```diff
@@ -1,6 +1,9 @@
+import cgi
```
yedpodtrzitko, line 1: `cgi` is deprecated since Python 3.8 and is removed in Python 3.10

kemingy:
* close this because of #225

PR #193, opened 2022-01-18
Fix blueprint find_routes method

guilhermelou:
## Not Finding Blueprint Routes

This PR fixes a small bug when finding blueprint routes. It's basically not finding any route if you use prefix.

### Code to reproduce the error:
```
from flask import Flask, jsonify, Blueprint
from spectree import Response, SpecTree

app = Flask(__name__)

blueprint_app = Blueprint("my_blueprint_app", __name__, url_prefix="/v1")
api = SpecTree("flask")


@blueprint_app.route("/my-blueprint-endpoint", methods=["GET"])
@api.validate(resp=Response("HTTP_200"), tags=["My Blueprint Endpoints"]
)
def my_blueprint_endpoint():
    return jsonify({}), 200


@app.route("/my-app-endpoint", methods=["GET"])
@api.validate(resp=Response("HTTP_200"), tags=["My App Endpoints"]
              )
def my_app_endpoint():
    return jsonify({}), 200


api.register(blueprint_app)
app.register_blueprint(blueprint_app)

```

### Print showing missing blueprint endpoints in the swagger docs, it shows only the app routes:
Screenshot 2022-01-18 at 22 23 26

I believe this was just a missing `not` in the `find_routes` method

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -21,7 +21,7 @@ class FlaskPlugin(BasePlugin):
             if (
                 self.blueprint_state
                 and self.blueprint_state.url_prefix
-                and str(rule).startswith(self.blueprint_state.url_prefix)
+                and not str(rule).startswith(self.blueprint_state.url_prefix)
             ):
                 continue
             yield rule
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -21,7 +21,12 @@ class FlaskPlugin(BasePlugin):
             if (
                 self.blueprint_state
                 and self.blueprint_state.url_prefix
-                and not str(rule).startswith(self.blueprint_state.url_prefix)
+                and (
+                    not str(rule).startswith(self.blueprint_state.url_prefix)
+                    or str(rule).startswith(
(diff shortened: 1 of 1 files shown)


spectree/plugins/flask_plugin.py
```diff
@@ -21,7 +21,7 @@ def find_routes(self):
             if (
                 self.blueprint_state
                 and self.blueprint_state.url_prefix
-                and str(rule).startswith(self.blueprint_state.url_prefix)
+                and not str(rule).startswith(self.blueprint_state.url_prefix)
```
kemingy, line 24:
```suggestion
                and (
                    not str(rule).startswith(self.blueprint_state.url_prefix)
                    or str(rule).startswith(
                        "/".join([self.blueprint_state.url_prefix, self.config.path])
                    )
                )
```

kemingy (commented):
Thanks for your contribution.

I think you need to filter the lambda functions.

spectree/plugins/flask_plugin.py
guilhermelou: Makes sense, committing suggestion.

kemingy (approved):

PR #199, opened 2022-02-09
Restored docstring paragraph formatting

### LINKED ISSUE #198: [BUG] Paragraph formatting is removed from openapi.json
When I annotate an operation in this way:

```def on_post(self, req, resp):
"""Title

paragraph 1

paragraph 2
"""
```

It actually ends up in openapi.json as `{"summary": "Title", "description": "paragraph 1 paragraph 2", ...}` and the entire description is shown on one line in both ReDoc and Swagger:

### Title

paragraph 1 paragraph 2

Obviously I'd like my paragraph formatting to be maintained. A workaround is adding `<br />` instead of newlines but that seems like a bunch of unnecessary clutter with possibly faulty assumptions on the way the spec is going to be displayed.

I have traced the cause of this problem to https://github.com/0b01001001/spectree/blob/a4f5321c946acbd27515988d4c44ba44ef63ab08/spectree/utils.py#L39
Here, all newlines are replaced by spaces. This was introduced in https://github.com/0b01001001/spectree/pull/175 and looking at the diff, it seems like a change in behaviour compared to before that PR, while it did not state addressing this for any reason. Commenting out that line correctly (IMHO) leaves the newlines in and the result becomes `{"summary": "Title", "description": "paragraph 1\n\nparagraph 2", ...}`, which correctly renders my desired formatting in both ReDoc and Swagger.

It seems to be an attempt at replacing newlines (purely those due to maximum line length) by spaces so single line breaks in the comment don't show up in the resulting description, however that has the side effect of removing all intentional formatting using newlines. Probably a better way to go around this is splitting the description into paragraphs where they are broken by multiple newlines, then replacing the remaining single newlines by spaces in each paragraph, and finally joining the result again. I have attached a patch that does this and works fine for me.

[docstring_paragraphs.patch.txt](https://github.com/0b01001001/spectree/files/8032292/docstring_paragraphs.patch.txt) -- edit: revised patch in https://github.com/0b01001001/spectree/issues/198#issuecomment-1033977452

DataGhost:
Fixes #198

### DIFF
--- a/spectree/utils.py
+++ b/spectree/utils.py
@@ -33,15 +33,14 @@ def parse_comments(func: Callable[..., Any]) -> Tuple[Optional[str], Optional[st
 
     docstring = re.split("\f", docstring, maxsplit=1)[0]
 
-    docstring_parts = re.split(r"\n\s*\n", docstring, maxsplit=1)
+    docstring_parts = re.split(r"\n\s*\n", docstring)
     for i in range(len(docstring_parts)):
-        docstring_parts[i] = docstring_parts[i].strip()
-        docstring_parts[i] = re.sub("\n+", " ", docstring_parts[i])
+        docstring_parts[i] = docstring_parts[i].replace("\n", " ")
 
     summary = docstring_parts[0]
     description = None
     if len(docstring_parts) > 1:
-        description = docstring_parts[1]
+        description = "\n\n".join(docstring_parts[1:])
 
     return summary, description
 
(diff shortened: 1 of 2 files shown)


DataGhost:
I forgot to run the tests and add some new ones. Unsure how extensive these tests need to be but I think this caught the most obvious cases. Since `cleandoc()` only cleans the left side of strings, I changed the original `strip()` call to just `rstrip()`, that should do the trick and also leaves the original test results as they were.

kemingy (approved):

PR #214, opened 2022-04-23
misc: use hash suffix instead of prefix in get_model_key

### LINKED ISSUE #213: [feature] easier way to override `get_model_key()`
Hi,
the function [utils.get_model_key](https://github.com/0b01001001/spectree/blob/9147598a03d12223fe3cdf1523d08d2390475e9c/spectree/utils.py#L220) is used for generating model IDs + their references in `openapi.json`. We try to use this as a source for generating matching models on a frontend (via `@rtk-query/codegen-openapi` ), but the model IDs are prefixed with sha1 hash (which can start with a number), and the mentioned tool cant handle that. 

I overrided `SpecTree._generate_spec()` where I manually inject a prefix everywhere, but that's quite inelegant, and the generated model names are unpleasant. 

Easy way out of this would be having `get_model_key()` defined in a way which could be overriden easily. Not sure how exactly that would be done, so I'm opening this issue for now to hear some feedback if I'm not missing something obvious here.

Thanks.

yedpodtrzitko:
closes #213

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -4,6 +4,7 @@ from typing import Dict, List
 from pydantic import BaseModel, Field, root_validator
 
 from spectree import SecurityScheme, Tag
+from spectree.utils import hash_module_path
 
 api_tag = Tag(name="API", description="🐱", externalDocs={"url": "https://pypi.org"})
 
@@ -133,3 +134,18 @@ WRONG_SECURITY_SCHEMAS_DATA = [
     },
     {"name": "wrong_Data", "data": {"x": "y"}},
 ]
+
+
+def get_model_path_key(model_path: str) -> str:
+    """
+    generate short hashed prefix for module path (instead of its path to avoid
+    code-structure leaking)
+
+    :param model_path: `str` model path in string
+    """
+
+    model_path, _, model_name = model_path.rpartition(".")
+    if not model_path:
+        return model_name
+
+    return f"{model_name}.{hash_module_path(module_path=model_path)}"
--- a/spectree/utils.py
+++ b/spectree/utils.py
@@ -199,34 +199,16 @@ def hash_module_path(module_path: str):
     return sha1(module_path.encode()).hexdigest()[:7]
 
 
-def get_model_path_key(model_path: str):
-    """
-    generate short hashed prefix for module path (instead of its path to avoid
-    code-structure leaking)
-
-    :param modelpath: `str` model path in string
-    """
-
-    model_path_parts = model_path.rsplit(".", 1)
-    if len(model_path_parts) > 1:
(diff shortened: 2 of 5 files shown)


yedpodtrzitko:
ok I actually ran the testsuite this time, since it's faster than waiting for the pipeline approval. I noticed one thing though - `get_model_path_key()` (which was making the tests fail) is being used only in the testsuite and never in the actual production code. As such, shouldnt be its definition moved into the testusite as well? Or maybe replaced with usage of `get_model_key()`.

kemingy:
> ok I actually ran the testsuite this time before pushing. I noticed one thing though - `get_model_path_key()` (which was making the tests fail) is being used only in the testsuite and never in the actual production code. As such, shouldnt be its definition moved into the testusite as well? Or maybe replaced with usage of `get_model_key()`.

Yeah. I didn't notice this before. I think we can move it to the tests part.

tests/common.py
```diff
@@ -133,3 +134,18 @@ def get_paths(spec):
     },
     {"name": "wrong_Data", "data": {"x": "y"}},
 ]
+
+
+def get_model_path_key(model_path: str) -> str:
+    """
+    generate short hashed prefix for module path (instead of its path to avoid
+    code-structure leaking)
```
kemingy, line 142:
```suggestion
    generate short hashed suffix for module path (instead of its path to avoid
    code-structure leaking)
```

Makefile
```diff
@@ -6,7 +6,7 @@ install:
 	pip install -e .[flask,falcon,starlette,dev]
 
 test:
-	pip install falcon --upgrade
+	pip install falcon starlette --upgrade
```
kemingy, line 9:
```suggestion
	pip install falcon --upgrade
```
The reason I put "falcon" here is that falcon2 is incompatible with falcon3. So I need to run tests for both of them.

I think starlette doesn't have this issue?

yedpodtrzitko: yes I assumed it's to test multiple Falcon's versions. The problem I stumbled upon was that the tests require Starlette to be installed, otherwise it fails with `ImportError`, so I hijacked this to install starlette as needed.

kemingy: I guess you mean when you run the `make test` in your machine, the "starlette" is required?

Actually, you can run `make install`. So all the required libraries will be installed.

CI:
* make install
* make lint
* make test

yedpodtrzitko: > Actually, you can run make install

that's fair, I'll revert this then

kemingy (approved):

kemingy (approved):

yedpodtrzitko:
wait a second with the potential merge please

yedpodtrzitko:
ok it's good to go now on my end

kemingy (approved):

PR #216, opened 2022-04-24
fix: handle case when no view response is defined

yedpodtrzitko:
hi,
there seems to be a regression introduced in #212

it can happen in two situations, each ending with different error:

-  the value of specific status code for `resp` in `@api.validate` is `None`:
```
>       if isinstance(model, resp.find_model(status)):
E       TypeError: isinstance() arg 2 must be a type or tuple of types
```

- `resp` in `@api.validate()` is not declared at all:
```
>       isinstance(model, resp.find_model(status)):
E       AttributeError: 'NoneType' object has no attribute 'find_model'
```

the error was raised in plugins for `flask` and `falcon`, `starlette` was working fine.

this PR adds fix for the two plugins + tests for all three of them.

### DIFF
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -226,11 +226,11 @@ class FalconPlugin(BasePlugin):
         func(*args, **kwargs)
 
         if resp and resp.has_model():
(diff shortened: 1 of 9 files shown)


kemingy (approved):
Thanks

PR #222, opened 2022-04-30
support list resp with List[Model]

### LINKED ISSUE #96: [Feature]Can't support list type swagger response

kemingy:
close #96

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,6 @@ lint:
 	isort --check --diff --project=spectree ${SOURCE_FILES}
 	black --check --diff ${SOURCE_FILES}
 	flake8 ${SOURCE_FILES} --count --show-source --statistics
-	mypy --install-types --non-interactive spectree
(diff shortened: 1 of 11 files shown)


kemingy:
Example:

```python
from typing import List, Type
from flask import Flask, jsonify
from spectree import SpecTree, Response


class User(BaseModel):
    name: str
    page: int


app = Flask(__name__)
spec = SpecTree("flask")


@app.route("/")
@spec.validate(resp=Response(HTTP_200=User))
def index():
    return User(name="John", page=1)


@app.route("/list")
@spec.validate(resp=Response(HTTP_200=List[User]))
def list_user():
    return jsonify([{"name": "John", "page": 1}, {"name": "Jane", "page": 2}])


if __name__ == "__main__":
    spec.register(app)
    app.run()
```

PR #224, opened 2022-05-27
Flask: add trailing slash to apidoc routes

yedpodtrzitko:
## Current behaviour

`add_url_rule()` in Flask plugin are defined without trailing slash, so only the paths without it work, ie:
 ```
$ curl -i http://localhost:5000/apidoc/swagger/ |grep HTTP
HTTP/1.0 404 NOT FOUND

$ curl -i http://localhost:5000/apidoc/swagger |grep HTTP
HTTP/1.0 200 OK
```

## Expected behaviour

Default behaviour in Flask regarding trailing slashes is [documented as follows](https://werkzeug.palletsprojects.com/en/2.1.x/routing/#rule-format):

> Rules that end with a slash are “branches”, others are “leaves”. If `strict_slashes` is enabled (the default), visiting a branch URL without a trailing slash will redirect to the URL with a slash 

adding the trailing slash into the routes definition makes both options work:
 ```
$ curl -i http://localhost:5000/apidoc/swagger/ |grep HTTP
HTTP/1.0 200 OK

$ curl -i http://localhost:5000/apidoc/swagger |grep HTTP
HTTP/1.0 308 PERMANENT REDIRECT
```

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -257,7 +257,7 @@ class FlaskPlugin(BasePlugin):
 
             for ui in self.config.page_templates:
                 app.add_url_rule(
-                    rule=f"/{self.config.path}/{ui}",
+                    rule=f"/{self.config.path}/{ui}/",
                     endpoint=f"openapi_{self.config.path}_{ui.replace('.', '_')}",
                     view_func=lambda ui=ui: gen_doc_page(ui),
                 )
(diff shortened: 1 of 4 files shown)


kemingy:
One more thing is the `/apidoc/openapi.json`. Will it be better to change this one?

---

BTW, I found falcon has another solution: [falcon/faq/trailing-slash](https://falcon.readthedocs.io/en/stable/user/faq.html?highlight=slash#how-does-falcon-handle-a-trailing-slash-in-the-request-path).

And starlette can handle it.

I guess no need to change the other two frameworks.

yedpodtrzitko:
> One more thing is the `/apidoc/openapi.json`. Will it be better to change this one?

I'd keep it as it is - you dont usually add slash after a name of file:  `example.com/foo.jpg/`
 feels weird in comparison with `example.com/foo.jpg`

kemingy (approved):
LGTM. Thanks for your contribution.

PR #225, opened 2022-05-28
add support for form data request 

yedpodtrzitko:
taking over where #185 was left. This is still very much WIP

- the previous PR made `form` and `json` exclusive (ie. only one can be defined at the time), which I dont think it should be enforced. Here's a use-case where we have right now: we have `/login/` endpoint, which can be called either from a legacy web which submits `<form>`, and it can be also used on a new frontend, which submits JSON payload. As such, one route can have both `form` and `json`.

- previous PR is adding usage of `cgi`, which is about to be deprecated, so it is again removed here

### DIFF
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -1,9 +1,10 @@
-import inspect
-import re
 import cgi
-from io import BytesIO
+import inspect
+import io
+import re
 from functools import partial
 from typing import Any, Callable, Dict, List, Mapping, Optional
+from io import BytesIO
 
 from pydantic import ValidationError
 
@@ -177,12 +178,12 @@ class FalconPlugin(BasePlugin):
             return [self.parse_field(subfield) for subfield in field]
 
         # When file name isn't ascii FieldStorage will not consider it.
-        encoded = field.disposition_options.get('filename*')
+        encoded = field.disposition_options.get("filename*")
         if encoded:
             encoding, filename = encoded.split("''")
             field.filename = filename
             field.file = BytesIO(field.file.read().encode(encoding))
-        if getattr(field, 'filename', False):
+        if getattr(field, "filename", False):
             return field
 
         # This is not a file, thus get flat value (not FieldStorage instance).
@@ -205,7 +206,8 @@ class FalconPlugin(BasePlugin):
             req.context.json = json.parse_obj(media)
         elif form:
             # This must be done to avoid a bug in cgi.FieldStorage.
-            req.env.setdefault('QUERY_STRING', '')
+            # https://gist.github.com/kgriffs/fe371bfbb2d63211889b
+            req.env.setdefault("QUERY_STRING", "")
 
             try:
                 _form = cgi.FieldStorage(fp=req.stream, environ=req.env)  # TODO: rename
@@ -284,22 +286,6 @@ class FalconAsgiPlugin(FalconPlugin):
     OPEN_API_ROUTE_CLASS = OpenAPIAsgi
     DOC_PAGE_ROUTE_CLASS = DocPageAsgi
 
-    async def async_parse_field(self, field):
-        if isinstance(field, list):
-            return [self.parse_field(subfield) for subfield in field]
-
-        # When file name isn't ascii FieldStorage will not consider it.
-        encoded = field.disposition_options.get('filename*')
-        if encoded:
-            encoding, filename = encoded.split("''")
-            field.filename = filename
-            field.file = BytesIO(await field.file.read().encode(encoding))
-        if getattr(field, 'filename', False):
-            return field
-
-        # This is not a file, thus get flat value (not FieldStorage instance).
-        return field.value
-
     async def request_validation(self, req, query, json, form, headers, cookies):
         if query:
             req.context.query = query.parse_obj(req.params)
@@ -317,16 +303,23 @@ class FalconAsgiPlugin(FalconPlugin):
             req.context.json = json.parse_obj(media)
         elif form:
             # This must be done to avoid a bug in cgi.FieldStorage.
-            req.scope.setdefault('QUERY_STRING', '')
+            req.scope.setdefault("QUERY_STRING", "")
+            # For WSGI compatibility
+            # https://asgi.readthedocs.io/en/latest/specs/www.html#wsgi-compatibility
+            req.scope.setdefault("REQUEST_METHOD", req.method)
 
             try:
-                _form = cgi.FieldStorage(fp=req.stream, environ=req.scope)  # TODO: rename
+                _form = cgi.FieldStorage(
+                    fp=io.BytesIO(await req.stream.read()),
+                    environ=req.scope,
+                    headers=req.headers,
+                )
             except ValueError:  # Invalid boundary?
                 raise
 
             req_form = {}
             for key in _form:
-                req_form[key] = await self.async_parse_field(_form[key])
+                req_form[key] = self.parse_field(_form[key])
 
             req.context.form = form.parse_obj(req_form)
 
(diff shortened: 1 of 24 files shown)


spectree/plugins/falcon_plugin.py
```diff
@@ -167,27 +169,51 @@ def parse_path(self, route, path_parameter_descriptions):
...
         if json:
+            try:
+                media = req.media
+            except self.FALCON_HTTP_ERROR as err:
+                if err.status not in self.FALCON_MEDIA_ERROR_CODE:
+                    raise
+                media = None
             req.context.json = json.parse_obj(media)
+        if form:
+            _form, files = multipart.parse_form_data(req.env)
```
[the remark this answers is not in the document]
yedpodtrzitko: will take look into it, thanks

spectree/plugins/falcon_plugin.py
```diff
@@ -175,27 +175,33 @@ def parse_path(self, route, path_parameter_descriptions):
...
+            except self.FALCON_HTTP_ERROR as err:
+                if err.status not in self.FALCON_MEDIA_ERROR_CODE:
+                    raise
+                media = None
             req.context.json = json.parse_obj(media)
+        if form:
+            # TODO - possible to pass the BodyPart here?
+            # req_form = {x.name: x for x in req.get_media()}
+            req_form = {x.name: x.stream.read() for x in req.get_media()}
+            req.context.form = form.parse_obj(req_form)
```
yedpodtrzitko, line 197: @vytas7 hi, you informed me previously about multipart forms support in Falcon here in this PR. I was looking into it, but I cant make it work in a way I'd be happy with. 

Here we're parsing the request data & passing it into Pydantic model for a validation, which is then making available in a view. I'd like to pass into Pydantic model directly the items returned from `get_media()` (is it `BodyPart` right?), but I think their content is getting evaluated while passing through Pydantic model, so when trying to read them again in the view later, it's returning just an empty bytes sequence. Which is what the docstring in `get_media()` says, but I dont know how to access the cached version of the data. Any advice please?

reading in view:
https://github.com/0b01001001/spectree/blob/d6c4e744552f8af3d719bba3511f0afbd7efd659/tests/test_plugin_falcon.py#L190-L193

spectree/plugins/falcon_plugin.py
```diff
@@ -271,12 +277,26 @@ async def request_validation(self, req, query, json, headers, cookies):
...
+                form_data = await req.get_media()
+            except self.FALCON_HTTP_ERROR as err:
+                if err.status not in self.FALCON_MEDIA_ERROR_CODE:
+                    raise
+                req.context.form = None
+            else:
+                res_data = {}
+                async for x in form_data:
+                    res_data[x.name] = x
+                    await x.data  # TODO - how to avoid this?
```
yedpodtrzitko, line 291: @vytas7 here in async version I dont know how to avoid passing the object through Pydantic model without awaiting it here. Any advice appreciated.

spectree/plugins/falcon_plugin.py
vytas7: Hi again @yedpodtrzitko!
Yes, Falcon's multipart support is designed for parsing the form on-the-fly, without buffering any data into the memory or temp files behind the user's back (because the former can make you run out of memory, and the latter kill performance, depending on what you're doing).

I'm not familiar with Spectree's codebase, what would you like to send into Pydantic, only the parts serialized as JSON(/YAML/msgpack/etc) and text fields, or also files? Would you like to spool file parts to tempfiles?

In general, if you want to pass a synchronous data structure into Pydantic, you'll need it to create a wrapper to store that data into when parsing.

yedpodtrzitko:
@kemingy how would you feel about making a breaking change and drop support for Falcon 2? It was released 2019 and never got any patch above 2.0.0 (see https://pypi.org/project/falcon/#history ), so it seems unsupported anyway.

The reason why I ask is it looks like it does not support form data. The pipeline right now fails because of `get_media()` does not exist in Falcon 2. But when I tried to use `req.media` instead, then I got the following error:
```
AssertionError: {"title": "Unsupported media type", "description": "multipart/form-data is an unsupported media type."}
```

kemingy:
Hi @yedpodtrzitko. I need to take some time to check whether it's possible to support this in falcon 2.

My concern is that falcon 2 will be the default one if users type `pip install falcon` between 2019-04-27 and 2021-04-06 (almost two years). There may be lots of legacy projects using falcon 2. If we are going to deprecate it, we'd better create a release with deprecate warning about falcon 2 before the final deprecation.

yedpodtrzitko:
> we'd better create a release with deprecate warning about falcon 2 before the final deprecation.

yes that's the standard practice and under normal circumstances I'd go that way too. However it seems like Falcon 2 is dead in the water - I haven't found any mention about its support anywhere and it didnt get any patch above 2.0.0. So if someone installed it in the timespan you mentioned and they havent upgraded by now, then Spectree's breaking change is the latest problem they could have.

Here's the docs for supported media types in [Falcon 2](https://falcon.readthedocs.io/en/2.0.0/api/media.html#media-type-constants
) and [Falcon 3](https://falcon.readthedocs.io/en/3.1.0/api/media.html#media-type-constants). v3 has `MEDIA_MULTIPART` there, v2 has nothing even remotely close to it.

So I can understand the appeal to do things The Right Way, but I dont think it's worth it in this case. If this will be a breaking change, the next version should be `1.0.0`, since breaking change should increase the major version number, right? And for the user who are really stuck with Falcon 2, the branch `0.10.x` can still receive bugfixes for a period of time.

kemingy:
> So I can understand the appeal to do things The Right Way, but I dont think it's worth it in this case. If this will be a breaking change, the next version should be `1.0.0`, since breaking change should increase the major version number, right? And for the user who are really stuck with Falcon 2, the branch `0.10.x` can still receive bugfixes for a period of time.

I see. I'm okay with this breaking change in `1.0.0`.

* Will review #239 ASAP

yedpodtrzitko:
> I see. I'm okay with this breaking change in `1.0.0`.

Ok. Do you feel ready (feature-wise) to have the project bumped to version 1.0? I assume that's how you'd do it anyway, but will there be at least one beta/rc release before having final v1.0.0?

kemingy:
> > I see. I'm okay with this breaking change in `1.0.0`.
> 
> Ok. Do you feel ready (feature-wise) to have the project bumped to version 1.0? I assume that's how you'd do it anyway, but will there be at least one beta/rc release before having final v1.0.0?

Yep. I think we can have an RC release before the final v1.0.0.

yedpodtrzitko:
Seems like Jinja 3.1+ has some breaking changes, so it needs to be pinned too. Probably in main branch instead of here.

Pypy tests seems to be broken, will take a look later.

yedpodtrzitko:
I had to revert some changes from #185 which were making PyPy pipeline fail, but it seems to be all passing now.

I would prefer to not squash the commits, so the work of folks who were working on this previously is properly attributed to them (I squashed it to one commit per each author already).

spectree/plugins/falcon_plugin.py
vytas7: @yedpodtrzitko It depends what you are trying to achieve. Do you want to buffer whole files in the memory? Or do you want to spool them to temp files like some other frameworks do?\

Falcon even puts a cap on a maximum amount of data that can be referenced this way (`await part.get_data()`) in order to avoid surprises such as running out of memory.

Use `await part.stream.read()` to read the whole part as a bytestring, or `await part.stream.pipe(async_file)`, or read by chunks, and store the result somewhere.

kemingy:
* So far, most of the code changes look good to me. I'm going to merge this PR(also the original #185) since it's opened for a very long time.
* Really appreciate your contributions @Michae1Weiss @yedpodtrzitko @vytas7 
* Also, I'll open another issue to keep track of the falcon form data read part. I'm not going to release a stable version until it's well handled.

kemingy (approved):

Issue #226, opened 2022-05-29
[BUG] `json` data required for GET request


yedpodtrzitko:
**Describe the bug**
`GET` requests cant provide data for `json` property of `validate()`, so it shouldnt be required/validated, but it seems to happen right now. See example below.

**To Reproduce**
consider following endpoint (in `tests/test_plugin_flask.py`). 
This view can handle `GET` request or `POST` request. `json` property is relevant only for `POST` request, however even `GET` request trigger its validation:
```
@app.route("/api/no_response", methods=["GET", "POST"])
@api.validate(
    json=Query,
)
def no_response():
    return {}
```

```
resp = client.get("/api/no_response")
assert resp.status_code == 200
```

**Error message**
```
AssertionError: b'[{"loc":["order"],"msg":"field required","type":"value_error.missing"}]\n
```

kemingy:
Hi @yedpodtrzitko, thanks for your contribution.

My colleagues tested positive for Covid-19 last weekend. I also had lunch with them. I don't feel very well but haven't confirmed yet. I may need to take a rest. Will try to review it as soon as possible.

yedpodtrzitko:
This sure can wait, I hope you'll get well soon 🤞

PR #227, opened 2022-05-29
refactoring: define flask tests in single place and reuse them

yedpodtrzitko:
Moreless identical tests for Flask plugin are written in three different files now (`test_plugin_flask.py`, `test_plugin_flask_view.py`, `test_plugin_flask_blueprint.py`). This PR put them into single places from where they can be imported into each respective test file.

This revealed that the `no_response` view was not tested properly everywhere (see #226), so until that will be resolved, the parameter `json=Query` for given view is commented out.

### DIFF
--- a/tests/flask_imports/__init__.py
+++ b/tests/flask_imports/__init__.py
@@ -1,0 +1,17 @@
+from .dry_plugin_flask import (
+    test_flask_doc,
+    test_flask_no_response,
+    test_flask_return_model,
+    test_flask_skip_validation,
+    test_flask_validate,
+    test_flask_validation_error_response_status_code,
+)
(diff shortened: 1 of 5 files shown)


yedpodtrzitko:
oops, missing rebase. brb

kemingy:
> This revealed that the `no_response` view was not tested properly everywhere (see #226), so until that will be resolved, the parameter `json=Query` for given view is commented out.

- Since #228 has been merged. This part can be updated accordingly.

Thanks for your contribution!

yedpodtrzitko:
rebased and updated

kemingy (approved):
LGTM

PR #228, opened 2022-06-07
fix: dont validate JSON data in request which cant provide it

### LINKED ISSUE #226: [BUG] `json` data required for GET request

yedpodtrzitko:
this should fix #226

- falcon plugin not affected, since there seems to be only class-based views with split GET and POST dispatchers

- flask tests not changed, as I hope #227 can be merged eventually

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -140,24 +140,28 @@ class FlaskPlugin(BasePlugin):
         req_cookies: werkzeug.datastructures.ImmutableMultiDict
         """
         req_query = get_multidict_items(request.args) or {}
-        if request.mimetype in self.FORM_MIMETYPE:
-            req_json = get_multidict_items(request.form) or {}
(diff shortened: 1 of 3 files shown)


kemingy (approved):
LGTM

Issue #229, opened 2022-06-13
logging in `default_after_handler` unusuble by default


yedpodtrzitko:
in `default_after_handler()` there's this logging for a validation error:

https://github.com/0b01001001/spectree/blob/ac3f20dac13f173d582ba7e2b6e015dcf0ae6fc3/spectree/utils.py#L193-L200

I might be missing something, but I dont see how this logging is useful unless I write my custom logging handler taking into account data from `extra`, or a custom `default_after_handler` where I call logging as needed.
I can understand that hardcoding the whole error into the logging message is also suboptimal, but it would feel like a more sane default where the logged value is actually useful revealing what exactly went wrong.

Also it should be most likely `logging.error` instead of `logging.info`. 

How do you handle information from this log yourself, if at all?

kemingy:
Good question.

There's a long history. In short:

1. some people don't like the `status code` in the HTTP header so they want to return everything with `status code == 200` and add the error messages in the HTTP body.
2. some people may need to record some metrics or add other error handling logic, so I provided the `before` and `after` handlers.

Can check the original PR:

* https://github.com/0b01001001/spectree/pull/52

Back to your question, this log itself is not very useful for users. It's just a demonstration. I usually define my own service level handlers. I agree that it's better to use `error` instead of `info`. Also, adding the information to the log content is better than adding to the extra part.

yedpodtrzitko:
fixed via #233

PR #233, opened 2022-06-24
fix: log the validation error in the default  handler

yedpodtrzitko:
this makes the logging in default `after` a bit more verbose, as discussed in #229


kemingy (approved):
LGTM

PR #236, opened 2022-07-12
Make email-validator an optional dependency

### LINKED ISSUE #235: Can the email validator dependency be optional?
`EmailStr` is used for the contact info, but the field is optional. It's a waste to install these additional dependencies if I don't plan on supplying a value for this field, and don't plan on using `EmailStr` for my own API's models. Or maybe I do want to supply an email in the contact info, but don't value validation of it enough to justify the additional dependencies.

Is there a nice way to make this dependency optional? Of course, it could just be an extra for spectree, but will that feel awkward or unintuitive for users?

MarkKoz:
Resolve #235

It can be installed with `pip install spectree[email]`. Otherwise, fall back to using a `str` as the type of the email field in the contact model of the config.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ setup(
     python_requires=">=3.6",
     install_requires=requires,
     extras_require={
+        "email": ["pydantic[email]>=1.2"],
         "flask": ["flask"],
         "falcon": ["falcon"],
         "starlette": ["starlette[full]"],
--- a/spectree/config.py
+++ b/spectree/config.py
@@ -1,11 +1,21 @@
 import warnings
 from enum import Enum
-from typing import Any, Dict, List, Mapping, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional
 
 from pydantic import AnyUrl, BaseModel, BaseSettings, EmailStr, root_validator
 
 from .models import SecurityScheme, Server
 from .page import DEFAULT_PAGE_TEMPLATES
+
+# Fall back to a str field if email-validator is not installed.
+if TYPE_CHECKING:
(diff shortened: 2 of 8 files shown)


spectree/config.py
```diff
@@ -27,7 +37,7 @@ class Contact(BaseModel):
     #: contact url
     url: Optional[AnyUrl] = None
     #: contact email address
-    email: Optional[EmailStr] = None
+    email: Optional[EmailFieldType] = None
```
MarkKoz, line 40: I wanted to make the docs show `str` for the type of the email field, but couldn't figure out how to override it besides the following:

1. Don't have `email-validator` installed while generating docstrings
2. Use [`autodoc_type_aliases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases), but that would require `from __future__ import annotations`

Thoughts? I'm not experienced with sphinx, so maybe there's a better way. In any case, it's not a big deal if it still shows as `EmailStr` I guess.

setup.py
```diff
@@ -42,6 +42,7 @@
     python_requires=">=3.6",
     install_requires=requires,
     extras_require={
+        "email": ["pydantic[email]>=1.2"],
```
MarkKoz, line 45: Unsure if this version constraint is redundant since it's already in `requirements.txt`.

kemingy: It's okay to have a redundant version constraint.

spectree/config.py
kemingy: I don't have a better solution. I'm okay with it.

kemingy (approved):
LGTM. Thanks!

PR #237, opened 2022-07-19
Possibility to disable endpoint + README additions of missing previous items

Vlczech:


yedpodtrzitko:
Hi, thanks for the PR.

I can see what are you aiming for, but I dont think this should be handled via a new parameter in `validate()`. Without increasing its complexity, the same result can be achieved by returning 404 eagerly in given endpoint, no?

Is there similar feature in any of the existing frameworks/libraries? I modified Spectree for my project quite a lot so showing without context is a bit pointless, but it boils down to using `if DEBUG: flask_app.add_url_rule('/my/dev/endpoint')`. Similarly in Django adding endpoint for is usually solved as `if settings.DEBUG: `urlpatterns += my_dev_url_patterns`


But that just my 5 cents, I'll leave it up to @kemingy

kemingy:
Appreciate your help to document `deprecated`.

As for the new feature of `accessible`, I cannot find this in the [OpenAPI spec](https://swagger.io/specification/) (correct me if I'm wrong).

Perhaps you can split them into two PR. We can discuss the `accessible` part.

yedpodtrzitko:
There seems to be no progress, feel free to reopen this PR when ready.

PR #238, opened 2022-07-22
fix: use mypy.ignore_missing_imports instead of suppressing them everywhere manually

yedpodtrzitko:
seems like these mypy errors are the last thing making the pipeline fail in #225. Instead of adding `type: ignore` to all newly added imports there, it would be better to disable these errors in one place. This PR is doing that.

`ignore_missing_imports` description in mypy docs:
https://mypy.readthedocs.io/en/stable/config_file.html#confval-ignore_missing_imports

- config options are moved into its config file instead of calling them manually

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -194,7 +194,8 @@ class FlaskPlugin(BasePlugin):
         before(request, response, req_validation_error, None)
         if req_validation_error:
             after(request, response, req_validation_error, None)
-            abort(response)  # type: ignore
+            assert response  # make mypy happy
+            abort(response)
 
         result = func(*args, **kwargs)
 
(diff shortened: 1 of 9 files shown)


spectree/plugins/flask_plugin.py
```diff
@@ -194,7 +194,8 @@ def validate(
         before(request, response, req_validation_error, None)
         if req_validation_error:
             after(request, response, req_validation_error, None)
-            abort(response)  # type: ignore
+            assert response  # make mypy happy
+            abort(response)
```
yedpodtrzitko, line 198: This was weird. Removing `type: ignore` here passed when running it on my machine, but it was raising error when running here. When I put it back, it was complaining about unused ignore. So the way how to oveercome it was to make sure the response is not `Optional`.

kemingy: Sometimes that happens even if the Python and mypy versions are the same. I'm not sure about the root cause.

PR #239, opened 2022-07-23
drop support for Falcon 2

yedpodtrzitko:
moreless copy-paste of my comment here: https://github.com/0b01001001/spectree/pull/225#issuecomment-1192452387

Falcon 2 was released 2019 and never got any patch above 2.0.0 (see https://pypi.org/project/falcon/#history ), so it seems unsupported anyway.

The reason why I would propse doing this is that Falcon 2 does not support form data. The pipeline in #225 fails right now because `get_media()` does not exist in Falcon 2, so I tried to use `req.media` instead, and I got the following error (with Falcon 2)
```
AssertionError: {"title": "Unsupported media type", "description": "multipart/form-data is an unsupported media type."}
```

Dropping Falcon 2 would then make implementing the form data support easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Yet another library to generate OpenAPI documents and validate requests & respon
 * Validate query, JSON data, response data with [pydantic](https://github.com/samuelcolvin/pydantic/) :wink:
 * Current support:
   * Flask [demo](#flask)
-  * Falcon [demo](#falcon) (including ASGI under Falcon 3+)
+  * Falcon [demo](#falcon) (including ASGI)
   * Starlette [demo](#starlette)
 
 ## Quick Start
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -16,7 +16,7 @@ Yet another library to generate OpenAPI documents and validate requests & respon
 * Validate query, JSON data, response data with [pydantic](https://github.com/samuelcolvin/pydantic/) :wink:
 * Current support:
   * Flask [demo](#flask)
-  * Falcon [demo](#falcon) (including ASGI under Falcon 3+)
+  * Falcon [demo](#falcon) (including ASGI)
```
kemingy, line 19: I think we can delete (includeing ASGI) here since we requires falcon > 3.0.0

kemingy (commented):
Also, can you help to change the falcon demo in README?

Another thing related to falcon 2 is the https://github.com/0b01001001/spectree/blob/278f1bd197e4b7ddee967d1b18d7f652da923144/spectree/plugins/falcon_plugin.py#L61-L64

* It's related to #174

README.md
yedpodtrzitko: done

yedpodtrzitko:
it's rebased now, pipeline passing

kemingy (approved):

PR #244, opened 2022-07-31
add copy of `werkzeug.parse_rule` which is now marked as internal

yedpodtrzitko:
I did dig a bit into new Werkzeug and didnt figure out a way how to replace `parse_rule`, so it seems like the fastest way how to make it work is to keep own copy of `parse_rule`.
 
plus there's microoptimization in separate commit to do `bypass()` before some route parsing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     install_requires=requires,
     extras_require={
         "email": ["pydantic[email]>=1.2"],
-        "flask": ["flask", "werkzeug<2.2"],
+        "flask": ["flask"],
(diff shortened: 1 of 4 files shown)


kemingy (approved):

kemingy:
> I did dig a bit into new Werkzeug and didnt figure out a way how to replace `parse_rule`, so it seems like the fastest way how to make it work is to keep own copy of `parse_rule`.

You're right. After checking the diff of werkzeug, I feel it's almost impossible to re-use the function from their new implementation since the `Rule._parse_rule` doesn't expose the intermediate data.

I guess it's okay to copy their original function for now.

PR #250, opened 2022-08-05
Use typing.get_type_hints instead of .__annotations__

kageurufu:
https://docs.python.org/3/library/typing.html#typing.get_type_hints

This properly resolves forward annotations, and fixes annotations when
 code uses `from __future__ import annotations`


kemingy (approved):
Thanks for your contribution! :tada:

kageurufu:
Glad to help :)

PR #257, opened 2022-09-05
added possibility to set any func name to openapi docs for falcon ASGI/WSGI.

AndrewOdn:
added small changes to falcon plugin wsgi/asgi, because redoc docs look not good with standart falcon func naming like on_post / on_get. if it need any improvements in code style or others, tell it pls, it will be done.

### DIFF
--- a/spectree/spec.py
+++ b/spectree/spec.py
@@ -114,6 +114,7 @@ class SpecTree:
         query: Optional[ModelType] = None,
         json: Optional[ModelType] = None,
         form: Optional[ModelType] = None,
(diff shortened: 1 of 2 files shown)


kemingy:
The `func.__name__` will only be used when `summary` is empty.

So if you need a better "name" to display in swagger/redoc, you can try to add a summary to the function.

```python
class Index:
    @spec.validate()
    def on_post(self, req, resp):
        """
        Summary of this endpoint (this will be used as the name displayed in swagger/redoc)

        More descriptions are here.
        """
```

AndrewOdn:
Ohh, really? Thank you bro

Issue #259, opened 2022-09-13
[Feature request] Global validation error model override


dfkki123508:
It would be really cool to not only be able to override the `validation_error_status` globally but also the __validation error model__. e.g.

```python
spectree.spec.SpecTree(..., validation_error_model=MyPydanticErrorModel)
```

What do you think? I would be happy to contribute. 

BR

kemingy:
It has a global `before` function which will be called in `validate > before() > endpoint func > after() > response`.

```py
spec = SpecTree(before=your_func_to_modify_the_error_model)
```

Check the example: https://github.com/0b01001001/spectree/blob/bb96e101b10bc02daffa75a1b910a104811476e0/tests/test_plugin_flask.py#L26-L28

And the doc: https://0b01001001.github.io/spectree/utils.html#spectree.utils.default_before_handler

dfkki123508:
Yes, I saw that you can intercept the error handling with before and after. However, in the docs there will always display the standard error model, but I want to override it with my own error model. Technically it is possible if I add the validation error response to every endpoint, but this is a lot of redundant work, so overriding the standard error model globally would be nice.

kemingy:
> Technically it is possible if I add the validation error response to every endpoint, but this is a lot of redundant work, so overriding the standard error model globally would be nice.

You don't have to override one by one. You can define the global `before` handle and it will work on all the endpoints.

dfkki123508:
Sure, but I'm referring to the generated docs. There is no possibility to _globally_ override the standard validation error.

kemingy:
This is global:

> ```python
> spec = SpecTree(before=your_func_to_modify_the_error_model)
> ```

This is a specific function:

```python
spec.validate(before=your_func_to_modify_the_error_model)
```

dfkki123508:
No this is not what I mean :smile: Sorry, for the confusion. I try to explain in a different way: This [line](https://github.com/0b01001001/spectree/blob/master/spectree/spec.py#L224) adds the validation error to the responses which is later used in `spec._generate_spec`. So the api docs for every endpoint (not overriding 422 response) will look like this:
```yml
    responses:
      '422':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ValidationError.6a07bef'
```
So I want to tell spectree to use a custom model by default instead of the standard validation error.

Does this make it clearer? Thanks for your answers.

kemingy:
> For apps in my project this would also be useful!

[...]

PR #260, opened 2022-09-18
initial version (#259)

dfkki123508:
@kemingy re: https://github.com/0b01001001/spectree/issues/259


kemingy:
I guess what you want is to show a different ValidationError in the swagger/redoc page.

But the triggered error is still `pydantic.ValidationError` instead of the customized one.

May I know your use case for this feature?

kemingy (approved):
Thanks. I think https://github.com/0b01001001/spectree/issues/259#issuecomment-1318226246 makes sense.

PR #262, opened 2022-10-06
Dmitry/quart plugin

dima-engineer:


kemingy:
Thanks. Will review it this weekend.

spectree/plugins/quart_plugin.py
```diff
@@ -0,0 +1,216 @@
...
+        cookies: Optional[ModelType],
+        resp: Optional[Response],
+        before: Callable,
+        after: Callable,
+        validation_error_status: int,
+        skip_validation: bool,
+        *args: Any,
+        **kwargs: Any,
+    ):
+        from quart import abort, jsonify, make_response, request
```
kemingy, line 109: I see your point. Let me check if we can avoid this.

setup.cfg
```diff
@@ -13,6 +13,9 @@ max-complexity = 15
 [tool:isort]
 profile = black
 
+[tool:pytest]
+asyncio_mode = auto
+
```
kemingy, line 18: Is it required?

dima-engineer: A will check on UTests for the async framework you already have. I'm wondering how it's implemented without asyncio_mode. Probably because of this also python 3.6 checks failed.

dima-engineer:
@kemingy let's try run tests now. I changed tests a little now they should work without pytest-asyncio. I believe now they should be passed in python 3.6 as well

spectree/plugins/quart_plugin.py
dima-engineer: fixed

kemingy:
> @kemingy let's try run tests now. I changed tests a little now they should work without pytest-asyncio. I believe now they should be passed in python 3.6 as well

I checked the doc of [quart testing](https://quart.palletsprojects.com/en/latest/how_to_guides/testing.html). I found that it doesn't provide a convenient way to test the response. I understand that's why you need to use `pytest-asyncio`.

dima-engineer:
Yes, if it's required to avoid pytest-asyncio we can use asyncio.run inside each UTest instead. This is what I did in my last commit. Some kind of workaround.

kemingy:
Hi @dima-engineer, I create a new PR #266 to fix the import problem. So you should be able to import everything you need at the beginning of `quart_plugin.py`. Hope this can help you remove the duplicate part.

Let me know if you have any questions.

dima-engineer:
Hi @kemingy !
I'm trying to run `make import_test` but getting the following error:
```
Traceback (most recent call last):
  File "tests/import_module/test_quart_plugin.py", line 3, in <module>
    SpecTree("quart")
  File "/home/dmitry/Programming/my-projects/spectree/spectree/spec.py", line 75, in __init__
    module = import_module(plugin.name, plugin.package)
  File "/home/dmitry/.pyenv/versions/3.6.14/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/dmitry/Programming/my-projects/spectree/spectree/plugins/quart_plugin.py", line 9, in <module>
    from .flask_plugin import Context, FlaskPlugin
  File "/home/dmitry/Programming/my-projects/spectree/spectree/plugins/flask_plugin.py", line 3, in <module>
    from flask import Blueprint, abort, current_app, jsonify, make_response, request
ModuleNotFoundError: No module named 'flask'
make: *** [Makefile:10: import_test] Error 1
```

I believe this is because of `QuartPlugin` is inherited from `FlaskPlugin`

dima-engineer:
So, I assume I need to import somehow FlasgPlugin and inherit from it without reading other imports from `flask_plugin.py` file. In this case import of flask stuff in this file raises the error above

kemingy:
> So, I assume I need to import somehow FlasgPlugin and inherit from it without reading other imports from `flask_plugin.py` file. In this case import of flask stuff in this file raises the error above

I see. Things are different from the Falcon plugins.

I think you can keep the redundant code here. We can improve this later in another PR.

dima-engineer:
Do you mean I should not inherit from FlaskPlugin at all?

kemingy:
> Do you mean I should not inherit from FlaskPlugin at all?

Yeah. I think it's not an easy task to make it work with inheritance. We can improve this later.

dima-engineer:
I agree. Sounds good

kemingy (approved):
LGTM. Thanks for your help!

PR #269, opened 2022-10-21
AND/OR root security configuration

kageurufu:
We have multiple security options available to our whole API,
 and this allows us to set that. It also does basic validation
 of the `security` dict

Signed-off-by: Franklyn Tackitt <franklyn@tackitt.net>


kemingy (approved):
LGTM. Thanks.

PR #293, opened 2023-03-21
add test for top-level json list, use syrupy snapshot

yedpodtrzitko:
- fix for https://github.com/0b01001001/spectree/discussions/292
- adding support for `List[Model]` does not look straightforward, so defining `ListModel` is needed for now 
- also the specs is now compared as a whole snapshot via Syrupy instead of manual subset. I had to touch that test, as the list of endpoints there has changed. This way it feels more convenient than updating it manually every time.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -46,6 +46,6 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pytest
+          pip install pytest syrupy
       - name: Test
         run: make test
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ setup(
             "isort~=5.10",
             "autoflake>=1.4,<3.0",
             "mypy>=0.971",
+            "syrupy>=4.0.0",
         ],
     },
 )
--- a/tests/__snapshots__/test_plugin/test_plugin_spec[falcon][full_spec].json
+++ b/tests/__snapshots__/test_plugin/test_plugin_spec[falcon][full_spec].json
@@ -1,0 +1,805 @@
+{
+  "components": {
+    "schemas": {
+      "Cookies.7068f62": {
+        "properties": {
+          "pub": {
+            "title": "Pub",
+            "type": "string"
+          }
+        },
+        "required": [
+          "pub"
+        ],
+        "title": "Cookies",
+        "type": "object"
+      },
+      "FormFileUpload.7068f62": {
+        "properties": {
+          "file": {
+            "format": "binary",
+            "title": "File",
+            "type": "string"
+          }
+        },
+        "required": [
+          "file"
+        ],
+        "title": "FormFileUpload",
+        "type": "object"
+      },
+      "Headers.7068f62": {
+        "properties": {
+          "lang": {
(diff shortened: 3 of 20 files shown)


yedpodtrzitko:
Wait a second... I didnt change anything in the end, just added test. Am I stupid?

kemingy:
> Wait a second... I eventually removed all the changes after codeql/mypy complains and havent changed anything in the end, just added test. Am I stupid?

🤣 I thought this is the pre-PR that only replaces the snapshot test.

yedpodtrzitko:
@kemingy I have looked at it again with fresh eyes and I dont see any straightforward non-hacky way how to support `List[Model]` in request validation params. Here's the part where it falls flat: 
`request_validation()` eventually calls `json.parse_obj(request_json_data)` where `json` is the instance of model.
If `json` would be `List[Model]` instead of `Model` then it fails, because `List` does not have the method `parse_obj()` (beside other things). 
Ok you can try to call `gen_list_model()`, but what if someone will do `List[Dict[Model]]` or similar shenanigan? The same would need to be done for other validation params (ie. `form`, `headers`...). That's just the first error where it fails now, but I think there would be a few more down the line, like adding the generated models into the specs. 

To me it seems like the benefit of supporting this is not worth it when it can be done by defining the "`ListModel`".

kemingy:
Agree. I guess we don't need to support the List[Model] in the request for now. Unless the upstream pydantic has some support.

kemingy:
I'm still learning snapshot testing, so correct me if I'm wrong. I wonder how to guarantee that consistent results have what we need.

For example, ```assert api.spec == snapshot_json(name="full_spec")``` should be able to detect that the `api.spec` always returns the same content. But how can we know if it contains all the information like the original test case?

yedpodtrzitko:
> I'm still learning snapshot testing, so correct me if I'm wrong. I wonder how to guarantee that consistent results have what we need.

No problem. The first time you run the test you have to generate the snapshots (by adding `--snapshot-update` to pytest) and it will automatically create the snapshot files for all existing snapshot comparisons. In this case the `assert api.spec == snapshot_json()` will create a snapshot containing what `api.spec` returns. Then you run pytest again (without `--snapshot-update`) it will compare the snapshot with `api.spec`. If all data are static (ie. no timestamps or random IDs/hashes changing on each run) then it will pass. To see it works as expected you can either modify the snapshot file or the content of `api.spec` (eg. commenting out a random endpoint) and see it will fail with info what is the difference.
The next time the data will change (eg. a new endpoint is added or any parameters are changed/updated), it will fail, but you run pytest again with `--snapshot-update`, and it will update the snapshot with the current returned value, and you can diff it to see if the diff there is what you expect.
I find it pretty convenient, as this way you can catch various discrepancies which would go under the radar with manually asserting a subset of the data and it's less writing as it generates the snapshots for you.

> But how can we know if it contains all the information like the original test case?

The intermediate step I did was keeping all the original `assert` in place and adding `assert api.spec == snapshot_json(name="full_spec")` after it. That way I knew it still behaved as before, and after this assumption was confirmed I deleted all the original `assert` and kept just the new one in place.

kemingy (approved):
LGTM. Thanks.

Issue #300, opened 2023-04-17
Key appended to Schema Component


AndreasBBS:
When the openapi.json is generated by spectree the schema components have a key appended to them, for example: `Pet.acc87c7`; I'd like it to be generetated as `Pet`. Is there a way to disable the key appendage?

kemingy:
It's generated by https://github.com/0b01001001/spectree/blob/78edc00241006832f32a0e569e7a27844ca9aa38/spectree/utils.py#L235-L244

The purpose is to avoid the same class name defined in different files.

For now, there is no elegant way to disable it. You can patch this function. Or you may strip the suffix. Depends on your requirements.

AndreasBBS:
I understand I can fork but I was trying to avoid having to do that. I understand the goal of avoiding having the same class defined in different files.

My use case is that I'm using tools on my API clients to convert the openapi component schemes into types/classes (for example: [openapi-typescript](https://github.com/drwpow/openapi-typescript)). So is it your opinion that this tools should remove the prefix when naming the classes? I'm not sure this prefixing behavior is something defined in the OpenAPI specification so it might not be fair to expect libraries that consume OpenAPI specifications to be able to handle it.

Thanks for the quick reply. I really appreciate that!

kemingy:
I see your points. I think there is no spec for this.

One possible way to make it easier is moving the related functions to the `BasePlugin` static methods. So people can inherit and override part of the methods. We can do this step by step. If you're interested in this, you can refactor this `get_model_key` function.

Similar to https://github.com/0b01001001/spectree/pull/297#discussion_r1156158690

AndreasBBS:
I've been looking at the code to try to implement this refactor but I found some issues with moving it to BasePlugin static methods.

In `response.py` I don't seem to have access to the current plugin/backend like I do in the Spectree class:
https://github.com/0b01001001/spectree/blob/78edc00241006832f32a0e569e7a27844ca9aa38/spectree/response.py#L137

Any suggestion to what to do here?

AndreasBBS:
> I've been looking at the code to try to implement this refactor but I found some issues with moving it to BasePlugin static methods.
> 
> In `response.py` I don't seem to have access to the current plugin/backend like I do in the Spectree class:
> 
> https://github.com/0b01001001/spectree/blob/78edc00241006832f32a0e569e7a27844ca9aa38/spectree/response.py#L137
> 
> Any suggestion to what to do here?

[...]

PR #302, opened 2023-04-21
Refactor the get_model_key util function

AndreasBBS:
Following the discussion in https://github.com/0b01001001/spectree/issues/300 this refactor moves the function `get_model_key()` from `utils.py` to `BasePlugin` as a staticmethod.

This will allow custom plugin implementations to override the way the model key is generated as described by @kemingy [here](https://github.com/0b01001001/spectree/issues/300#issuecomment-1515620897).

The following example describes how to create a custom flask plugin that does not append the module hash to the model name:
```python
class YourFlaskPlugin(FlaskPlugin):
    def get_model_key(model: ModelType) -> str:
        return f"{model.__name__}"

spec = SpecTree(backend=YourFlaskPlugin)
```

EDIT:

> After lots of attempts to refactor through the `staticmethod` on the Base Plugin and continuously running into cyclical imports I decided to change course in the implementation. I decided to go with the good old Strategy pattern. Here's the new usage guide:
> 
> ```python
> spec = SpecTree(naming_strategy=lambda model: f"{model.__name__}")
> ```

### DIFF
--- a/spectree/_types.py
+++ b/spectree/_types.py
@@ -1,10 +1,22 @@
-from typing import Any, Dict, Iterator, List, Mapping, Optional, Sequence, Type, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Type,
+    Union,
+)
 
 from pydantic import BaseModel
 from typing_extensions import Protocol
 
 ModelType = Type[BaseModel]
 OptionalModelType = Optional[ModelType]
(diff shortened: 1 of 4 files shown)


AndreasBBS:
After lots of attempts to refactor through the `staticmethod` on the Base Plugin and continuously running into cyclical imports I decided to change course in the implementation.
I decided to go with the good old Strategy pattern. Here's the new usage guide:
```
spec = SpecTree(naming_strategy=lambda model: f"{model.__name__}")
```

kemingy (approved):

PR #305, opened 2023-04-24
Add Support for customizing how nested names are generated

AndreasBBS:
In the normal behavior the nested models are suffixed with the parent model name. This merge allows to override that behavior in the `Spectree` class. For example:

```python
spec = SpecTree(
    nested_naming_strategy=lambda _, child: child 
)
```


kemingy (approved):

PR #308, opened 2023-05-03
chore: clean test warnings

kemingy:


kemingy:
- py37 is no longer supported in flask: https://github.com/pallets/flask/pull/5072

Issue #248, opened 2022-08-03
Flask multiple request body examples


IlyaKalosha1:
Hi,
is there a way to set multiple examples for body when using spectree with flask.
The example how it's done in FastAPI - https://fastapi.tiangolo.com/tutorial/schema-extra-example/#body-with-multiple-examples:~:text=%3A-,Body,examples,-%C2%B6

Thanks

kemingy:
Hi @IlyaKalosha1, thanks for asking. I checked the [FastAPI examples details](https://fastapi.tiangolo.com/tutorial/schema-extra-example/#technical-details) and [OpenAPI spec](https://swagger.io/specification/). I found this is not supported yet.

Reason:

* FastAPI added the `examples` to `"requestBody"/"content"/"application/json"/examples` (can support multiple examples) rather than the normal path `"schemas"/"{SchemaName}"/example` (only support one)

Need to change some code to support this.

- [ ] support multiple examples for the request body

PR #309, opened 2023-05-03
Add Support for multiple examples in application/json

guiluko:
Add support for `examples` Object in OpenAPI.

### DIFF
--- a/spectree/utils.py
+++ b/spectree/utils.py
@@ -76,7 +76,14 @@ def parse_comments(func: Callable[..., Any]) -> Tuple[Optional[str], Optional[st
     return summary, description
 
 
-def parse_request(func: Any) -> Dict[str, Any]:
+def has_examples(schema_exta: dict) -> bool:
+    for _, v in schema_exta.items():
+        if isinstance(v, dict) and "value" in v.keys():
+            return True
+    return False
+
+
+def parse_request(func: Any, model: Optional[Any] = None) -> Dict[str, Any]:
     """
     get json spec
     """
@@ -85,6 +92,11 @@ def parse_request(func: Any) -> Dict[str, Any]:
         content_items["application/json"] = {
             "schema": {"$ref": f"#/components/schemas/{func.json}"}
         }
+
+        if model:
+            schema_extra = getattr(model.__config__, "schema_extra", None)
+            if schema_extra and has_examples(schema_extra):
+                content_items["application/json"]["examples"] = schema_extra
 
     if hasattr(func, "form"):
         content_items["multipart/form-data"] = {
--- a/spectree/response.py
+++ b/spectree/response.py
@@ -4,7 +4,7 @@ from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 from pydantic import BaseModel
 
 from ._types import ModelType, NamingStrategy, OptionalModelType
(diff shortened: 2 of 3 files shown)


yedpodtrzitko:
Hi, 
thanks for the PR. Can you please add a test to see how does this manifest? Thanks.

guiluko:
> Hi, thanks for the PR. Can you please add a test to see how does this manifest? Thanks.

Hi thanks for the recommandation, I added some.

For the context, I did this PR to close this issue https://github.com/0b01001001/spectree/issues/248

spectree/utils.py
```diff
@@ -76,7 +76,14 @@ def parse_comments(func: Callable[..., Any]) -> Tuple[Optional[str], Optional[st
     return summary, description
 
 
-def parse_request(func: Any) -> Dict[str, Any]:
+def has_examples(schema_exta: dict) -> bool:
+    for _, v in schema_exta.items():
```
yedpodtrzitko, line 80:
```suggestion
def has_examples(schema_extra: dict) -> bool:
    for _, v in schema_extra.items():
```

spectree/utils.py
```diff
@@ -76,7 +76,14 @@ def parse_comments(func: Callable[..., Any]) -> Tuple[Optional[str], Optional[st
...
+        if isinstance(v, dict) and "value" in v.keys():
```
yedpodtrzitko, line 81: nitpick: `.keys()` is superfluous since it's the default place where `in` is checking

```suggestion
        if isinstance(v, dict) and "value" in v:
```

spectree/utils.py
```diff
@@ -86,6 +93,11 @@ def parse_request(func: Any) -> Dict[str, Any]:
             "schema": {"$ref": f"#/components/schemas/{func.json}"}
         }
 
+        if model:
+            schema_extra = getattr(model.__config__, "schema_extra", None)
+            if schema_extra and has_examples(schema_extra):
+                content_items["application/json"]["examples"] = schema_extra
```
yedpodtrzitko, line 99: This code checks if there's ANY dictionary with a key named `value` in `schema_extra`, and based on that it assumes that ALL items in `schema_extra` are examples. Which might not be necessarily the case:

```
class Data(BaseModel):
    value: bool

    class Config:
        schema_extra = {
            "examples": {
                "example1": {"value": {"key1": "value1", "key2": "value2"}},
                "example2": {"value": {"key1": "value1", "key2": "value2"}},
            },
            "properties": {"value": {"this is not an example": True}},
        }
```

I think it could be better to nest all examples into `schema_extra.examples`, check presence of `examples`, and then then take just that as examples:

if has_examples(...):
    content_items["application/json"]["examples"] = schema_extra.examples

What do you think?

tests/test_response.py
```diff
@@ -107,6 +107,34 @@ def test_response_spec():
     assert spec.get(404) is None
 
 
+def test_response_spec_with_schema_extra():
```
yedpodtrzitko, line 110: if possible, can you use the fixture `snapshot_json` and compare the whole payload instead of cherrypicked parts? See example here:

https://github.com/0b01001001/spectree/blob/0508b707811a85e8294afa4a14fedd6b9b66ba57/tests/test_plugin.py#L25-L34


to generate the snapshot, add the following parameter: `pytest --snapshot-update`

PR #313, opened 2023-05-30
Extract the logic for getting the operation_id to a plugin method

Dekacc:
This is a continuation of the abandoned [#297 Allow overriding operation_id for each path on a single view function](https://github.com/0b01001001/spectree/pull/297) pull request.

I am a colleague of the author of that PR, and at our company we need to be able to override the implementation for the `operation_id` fields. Extracting that functionality in a separate method like this will enable us to do it.

I followed [this suggestion](https://github.com/0b01001001/spectree/pull/297#discussion_r1156158690) from @kemingy for this PR.


kemingy (approved):

PR #317, opened 2023-06-27
Validation with body which is intended for plain-text

Vlczech:


spectree/utils.py
```diff
@@ -91,6 +91,11 @@ def parse_request(func: Any) -> Dict[str, Any]:
             "schema": {"$ref": f"#/components/schemas/{func.form}"}
         }
 
+    if hasattr(func, "body"):
+        content_items["text/plain"] = {
+            "schema": {"$ref": f"#/components/schemas/{func.body}"}
+        }
```
kemingy, line 97: What will the BaseModel look like? Can you provide an example? If possible, can you add some tests?

Vlczech: @kemingy It is intended for text inputs like CSV etc. BaseModel is used to describe data in this model and it can be like this:

```
class FooBarPostBodyData(BaseModel):
    """
    CSV body for example for some import
    """
    __root__: str = Field(description="Data in CSV format - it consists of columns name, surname, ..., ..., ... .")
```

Vlczech: @kemingy : ?

kemingy: Sorry for the late response.

I feel that if we need to create a special argument for each [MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), there will be endless arguments. By the way, they all belong to the request body. I wander if we can add an attribute to the request model, for example, `spec_mime_type` or something else. This should be able to handle this kind of feature. What do you think?

kemingy: Some old discussions here https://github.com/0b01001001/spectree/discussions/176

Vlczech: Yes, it could be useful to process on one place for multiple mime-types.

ttekampe:
nice feature, I would also be interested in plain text / csv or also binary bodies.
So far the validator function has `json` and `form` arguments and I agree that adding one arg for every content type does not seem nice. So the way to go would be to add a `body` arg (probably deprecate `json`and `form`) and the body arg should be a class that specifies the content type and a model if necessary?

kemingy:
> nice feature, I would also be interested in plain text / csv or also binary bodies. So far the validator function has `json` and `form` arguments and I agree that adding one arg for every content type does not seem nice. So the way to go would be to add a `body` arg (probably deprecate `json`and `form`) and the body arg should be a class that specifies the content type and a model if necessary?

Yes. I think we can add the `body` and then gradually deprecate `json` and `form` in the future.

iongion:
Still any plans on integrating this ? 

There is this fork https://github.com/turner-townsend/flask-pydantic-spec which took this approach in its design.

My api supports application/bson as input format, along with application/json, that `json` seems unnatural.

kemingy:
> Still any plans on integrating this ?
> 
> There is this fork https://github.com/turner-townsend/flask-pydantic-spec which took this approach in its design.
> 
> My api supports application/bson as input format, along with application/json, that `json` seems unnatural.

Yes. I plan to support all kinds of MIME, which might introduce some breaking changes in v2.

Issue #307, opened 2023-04-24
feat: support `msgspec` as another schema definition


kemingy:
### Describe the feature

So far, we're using `pydantic` since:
1. validation
2. generate OpenAPI JSON schema

Recently, another library called `msgspec` has the same features with a great performance. (pydantic also rewrites the core part with Rust)

We should consider adding `msgspec` as another schema definition.

Refer to https://jcristharif.com/msgspec/jsonschema.html

kemingy:
duplicate with #329

PR #319, opened 2023-07-02
Upgrade to Pydantic v2

jaykv:
This will have to be tested more thoroughly of course, but I hope this is a good start to move towards pydantic v2 :)

Changes:
- New `BaseModel` method names, decorators, and config dict:
  * `model.parse_obj` -> `model.model_validate`
  * `model.dict` -> `model.model_dump(mode="json")`
  * `@root_validator` -> `@model_validator`
  * `@validator` -> `@field_validator`
  * `parse_raw(raw_stuff)` -> `model_validate(json.loads(raw_stuff))`
  * `schema` -> `model_json_schema`
  * `json` -> `model_dump_json`
  * `class Config: ...` -> `model_config = ConfigDict(...)`
- `BaseModel` subclasses using `__root__` switched to subclass `RootModel` with `root` annotation
- New fields for `ValidationErrorElement`: `input` and `url`
- Updated custom type definition for `BaseFile` ([thanks to flask-openapi3's implementation](https://github.com/luolingchun/flask-openapi3/blob/v3.x/flask_openapi3/models/file.py#L11))
- Response on `ValidationError` updated to use `err.json(indent=2)` instead of `jsonify(err.errors())` to avoid `TypeError: _ is not JSONSerializable` errors from the new `input` field (`indent=2` to keep response consistent to current)
- Update model schema definitions key from `definitions` -> `$defs`
- `req_validation_error.model.__name__` -> `req_validation_error.title`
- `pydantic.EmailStr.validate` -> `pydantic.validate_email`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,8 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    "pydantic>=1.9.1",
+    "pydantic>=2.0",
+    "pydantic-settings"
 ]
 
 [project.optional-dependencies]
@@ -37,7 +38,7 @@ dev = [
     "syrupy>=4.0",
 ]
 email = [
-    "pydantic[email]>=1.2,<2",
+    "pydantic[email]>=2.0",
 ]
 falcon = [
     "falcon>=3.0.0",
--- a/setup.py
+++ b/setup.py
@@ -6,10 +6,11 @@ from setuptools import setup
 setup(
     name="spectree",
     install_requires=[
-        "pydantic>=1.2",
+        "pydantic>=2.0",
+        "pydantic-settings"
     ],
     extras_require={
-        "email": ["pydantic[email]>=1.2,<2"],
+        "email": ["pydantic[email]>=2.0"],
         "flask": ["flask"],
         "quart": ["quart"],
         "falcon": ["falcon>=3.0.0"],
--- a/spectree/spec.py
+++ b/spectree/spec.py
@@ -75,7 +75,7 @@ class SpecTree:
         self.after = after
         self.validation_error_status = validation_error_status
         self.validation_error_model = validation_error_model or ValidationError
-        self.config: Configuration = Configuration.parse_obj(kwargs)
+        self.config: Configuration = Configuration.model_validate(kwargs)
         self.backend_name = backend_name
         if backend:
             self.backend = backend(self)
@@ -297,7 +297,7 @@ class SpecTree:
                 for tag in func_tags:
                     if str(tag) not in tags:
                         tags[str(tag)] = (
-                            tag.dict() if isinstance(tag, Tag) else {"name": tag}
+                            tag.model_dump() if isinstance(tag, Tag) else {"name": tag}
                         )
 
                 routes[path][method.lower()] = {
@@ -335,12 +335,12 @@ class SpecTree:
 
         if self.config.servers:
             spec["servers"] = [
-                server.dict(exclude_none=True) for server in self.config.servers
+                server.model_dump(exclude_none=True) for server in self.config.servers
             ]
 
         if self.config.security_schemes:
             spec["components"]["securitySchemes"] = {
-                scheme.name: scheme.data.dict(exclude_none=True, by_alias=True)
+                scheme.name: scheme.data.model_dump(exclude_none=True, by_alias=True)
                 for scheme in self.config.security_schemes
             }
 
@@ -353,11 +353,11 @@ class SpecTree:
         """
         definitions = {}
         for name, schema in self.models.items():
-            if "definitions" in schema:
-                for key, value in schema["definitions"].items():
+            if "$defs" in schema:
+                for key, value in schema["$defs"].items():
                     composed_key = self.nested_naming_strategy(name, key)
                     if composed_key not in definitions:
                         definitions[composed_key] = value
-                del schema["definitions"]
+                del schema["$defs"]
 
         return definitions
--- a/spectree/spec.py
+++ b/spectree/spec.py
@@ -335,12 +335,15 @@ class SpecTree:
 
         if self.config.servers:
             spec["servers"] = [
-                server.model_dump(exclude_none=True) for server in self.config.servers
+                server.model_dump(exclude_none=True, mode="json")
+                for server in self.config.servers
             ]
 
         if self.config.security_schemes:
             spec["components"]["securitySchemes"] = {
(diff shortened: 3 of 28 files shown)


pyproject.toml
```diff
@@ -22,7 +22,8 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    "pydantic>=1.9.1",
+    "pydantic>=2.0",
+    "pydantic-settings"
```
yedpodtrzitko, line 26:
```suggestion
    "pydantic-settings",
```

setup.py
```diff
@@ -6,10 +6,11 @@
 setup(
     name="spectree",
     install_requires=[
-        "pydantic>=1.2",
+        "pydantic>=2.0",
+        "pydantic-settings"
```
yedpodtrzitko, line 10:
```suggestion
        "pydantic-settings",
```

kemingy (commented):
Thanks for your contribution.

Pydantic v2 has too many breaking changes that lots of 3rd libraries are not able to upgrade in a short time.

This PR is not backward compatible (due to pydantic breaking changes). I'm not sure if we should make this library compatible with both pydantic v1 & v2. Or maybe we should add another abstract layer so that we can make it possible to support pydantic v1 & v2 and msgspec (ref #307)

spectree/spec.py
```diff
@@ -335,12 +335,15 @@ def _generate_spec(self) -> Dict[str, Any]:
 
         if self.config.servers:
             spec["servers"] = [
-                server.dict(exclude_none=True) for server in self.config.servers
+                server.model_dump(exclude_none=True, mode="json")
```
yedpodtrzitko, line 338: previously `.dict()` was returning a Python dictionary, now `.model_dump(mode="json")` returns json. Is this change intentional and if so, what difference does it make (I'm not familiar with this part of codebase, so I'd really like to know)?

tests/test_plugin_starlette.py
```diff
@@ -365,7 +365,7 @@ def test_starlette_no_response(client):
     resp = client.get("/api/no_response")
     assert resp.status_code == 200, resp.text
 
-    resp = client.post("/api/no_response", json={"name": "starlette", "limit": 1})
+    resp = client.post("/api/no_response", json={"name": "starlette", "limit": "1"})
```
yedpodtrzitko, line 368: why has this changed?

yedpodtrzitko:
That's a huge contribution, your effort is really appreciated. 

Pydantic v2 however still have some bugs and the ecosystem hasnt adapted yet, so I think it's too early to merge this at this moment.

spectree/spec.py
jaykv: `mode="json"` just serializes the values for the returned dict to make sure there's no python objects so when it's jsonified for `openapi.json`, it won't cause any `_ is not JSONSerializable` errors to pop up when accessing apidocs. I noticed it happening with some openapi config models in general when running the falcon_demo.

in this case though and probably elsewhere, I might have overused it since `Server` model doesn't have any special types

tests/test_plugin_starlette.py
jaykv: Yeah this one was weird. So `/api/no_response` has this decorator `@api.validate(json=StrDict, resp=Response(HTTP_200=None),)` and `StrDict` model has annotation `root: Dict[str, str]`. With v1, when we passed `"limit": 1`, `parse_obj` would internally coerce that into a string, but now `model_validate` is more strict about coercing it seems like

jaykv:
I understand of course, this is definitely just here to get the conversation started and test the waters. 

So is the ideal separation of layers for future spectree this?
1. Request/response model validation (pydantic v1/v2, msgspec)
2. OpenAPI schema generation
3. Route parsing, generation and validation integration into backends (flask, falcon, starlette, quart)
4. Apidocs (swagger, redoc)

kemingy:
> I understand of course, this is definitely just here to get the conversation started and test the waters.
> 
> So is the ideal separation of layers for future spectree this?
> 
>     1. Request/response model validation (pydantic v1/v2, msgspec)
> 
>     2. OpenAPI schema generation
> 
>     3. Route parsing, generation and validation integration into backends (flask, falcon, starlette, quart)
> 
>     4. Apidocs (swagger, redoc)

Yeah. I wonder if we can detect the model (pydantic v1/v2, megspec) automatically. This layer should be used to validate the request/response and generate the OpenAPI model schema.

AndreiDrang:
Hi
Perhaps it makes sense to just make a separate library? Smth like `spectreeV2`?

kemingy:
> Hi Perhaps it makes sense to just make a separate library? Smth like `spectreeV2`?

You can use `pip install git+https://github.com/0b01001001/spectree.git@refs/pull/319/head` directly.

DanCardin:
At least for v2, pydantic appears to be including a `v1` backport under `pydantic.v1` inside the v2 versions. That may or may not be useful in terms of converting internal usage or making it cross-compatible.

But it **seems** like it might make it possible for the library at current moment to "simply" alias all of its imports to pydantic to the v1 version. Such that I, as a user of pydantic (that doesn't necessarily stop at API models) could upgrade to v2 without impacting the spectree-compatible v1 API models.

kemingy:
> At least for v2, pydantic appears to be including a `v1` backport under `pydantic.v1` inside the v2 versions. That may or may not be useful in terms of converting internal usage or making it cross-compatible.
> 
> But it **seems** like it might make it possible for the library at current moment to "simply" alias all of its imports to pydantic to the v1 version. Such that I, as a user of pydantic (that doesn't necessarily stop at API models) could upgrade to v2 without impacting the spectree-compatible v1 API models.

This sounds like a good temporary solution.

jaykv:
> At least for v2, pydantic appears to be including a `v1` backport under `pydantic.v1` inside the v2 versions. That may or may not be useful in terms of converting internal usage or making it cross-compatible.
> 
> But it **seems** like it might make it possible for the library at current moment to "simply" alias all of its imports to pydantic to the v1 version. Such that I, as a user of pydantic (that doesn't necessarily stop at API models) could upgrade to v2 without impacting the spectree-compatible v1 API models.

I've seen this approach used specifically with just:
```python
try:
    from pydantic import v1 as pydantic
except ImportError:
    import pydantic
```

I didn't explore this yet, but could be worth looking into since it's a smaller change and also less likely to break for current v1 users. The only downside is spectree internally would still be using the v1 API instead of the v2 API. Although, that might not even matter since the long-term plan is to rehaul the internals to support msgspec anyways.

#324 is exploring the reverse approach for fun and upgrading the code to v2 API and creating backports for v1 and version-specific logic to handle the breaking changes.

PR #324, opened 2023-07-09
Support pydantic v1 and v2 with shims

jaykv:
This super-unpolished draft PR is mainly an exploration to support both v1 and v2 pydantic based on the discussion from #319 and how some other libraries (notably [FastAPI](https://github.com/tiangolo/fastapi/blob/c165be380f87b7b079f98c5134cccef0c9d67e4d/fastapi/_compat.py#L46-L494) and [app-model](https://github.com/pyapp-kit/app-model/blob/main/src/app_model/_pydantic_compat.py)) in the ecosystem are handling the pydantic v2 upgrade.

Some cons with this approach:
- Messy and feels hacky
- Complex maintainability

With that in mind, this could still serve as a decent temporary solution to provide flexibility. For long-term, I think all the internals will need to be rehashed to support pydantic v1/v2 and other validator libraries like msgspec in a more structured way.

Quick overview of the changes:
- `_pydantic.py` holds all the shims and compat things for pydantic v1 and v2
- All the internal code imports from `_pydantic` and follows v2 pydantic API when possible (for v1, that works either through the shims or the aliases). When that is not possible, there are if/else clauses to separate v1 and v2 convention as necessary (mainly inside the models and some test logic)
- Testing setup to run unit and snapshot tests with pydantic v1 and v2
- Passes linting/tests locally on 3.11.2, untested on older versions


jaykv:
Closing in favor of #326

Issue #329, opened 2023-07-27
feat: add an abstract layer to support pydantic/msgspec


kemingy:
Pydantic v2 has too many breaking changes that lots of 3rd libraries are not able to upgrade in a short time.

This PR is not backward compatible (due to pydantic breaking changes). I'm not sure if we should make this library compatible with both pydantic v1 & v2. Or maybe we should add another abstract layer so that we can make it possible to support pydantic v1 & v2 and msgspec (ref #307)

_Originally posted by @kemingy in https://github.com/0b01001001/spectree/pull/319#pullrequestreview-1510211443_

BradLucky:
Is this issue/repo still getting attention? This is over a year old at this point. I managed to copy `GenericModel` from Pydantic, making it so that SpecTree can still be used to depend on Pydantic V1 even though we are running V2. SpecTree has been great for us, but if it will not be getting true support for V2, I'm afraid I'll have to remove the dependency and find another solution.

kemingy:
#387 makes it possible to support both pydantic v1 & v2.

alexted:
@kemingy Thank you so much for taking the time and making the effort to look into this matter

alexted:
Hi! I understand you've decided to break the work down into several stages, two of which have already been completed. 
Are there still a few more steps to go?

kemingy:
> Hi! I understand you've decided to break the work down into several stages, two of which have already been completed. Are there still a few more steps to go?

Yeah, the next step is to introduce `msgspec` to see if there are any missing parts need to be abstracted. I'm already working on this one.

kemingy:
Hi @alexted, the PR #479 can pass some basic msgspec tests. Can you give it a try? Looking forward to your feedback!

alexted:
Sorry for the delayed response, it took some time. Thank you for your work - it looks great and works like magic! 
The adapter is a solid foundation. I left you a detailed comment in the PR.

alexted:
@kemingy  

- Could you clarify whether it will now be possible to create custom adapters for other serializers/validators, or if the current adapter abstraction doesn’t support that?

- This new functionality seems like it would warrant a new major version, doesn’t it, huh?

kemingy:
> [@kemingy](https://github.com/kemingy) Could you clarify whether it will now be possible to create custom adapters for other serializers/validators, or if the current adapter abstraction doesn’t support that? This new functionality seems like it would warrant a new major version, doesn’t it, huh?

[...]

PR #335, opened 2023-08-04
Support returning lists of unserialized models

jean-edouard-boulanger:
## Description

Currently, `spectree` only partially supports validation of list responses. For example:

```python
class User(BaseModel):
    user_id: str

@app.route()
@api.validate(
    resp=Response(HTTP_200=List[User])
)
def get_users():
    return [
        User(user_id="user1").dict(),  # pre-serialization of individual entries is required, or spectree crashes
        User(user_id="user2").dict()
    ]
```

You'll notice in particular that individual entries **must** currently be serialized before returning or `spectree` crashes. This is inconsistent with the rest of the interface, because it's perfectly okay to return models (without serializing them) - so why not list of models?

The change basically updates individual plugins to serialize individual list entries (when needed) before the final response is created. The change is conservative and the above logic is only applied when all of the below applies:

- Routes / responses that expect a list result.
- Result is indeed a list.
- Individual list entry is a `BaseModel` subclass.

I have not changed any of the existing validation logic in this case: we still rely on the generated root model for validation.

Change summary:

- Update the validation logic to support returning lists of models from endpoints (for all plugins). I have made sure that returning lists of serialized models remains supported.
- Small typing update so this `Response(HTTP_200=List[User])` no longer triggers IDEs / type validation frameworks.
- Added tests for this use case (for all plugins).

### DIFF
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -6,7 +6,7 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, get_type_hints
 from falcon import HTTP_400, HTTP_415, HTTPError
 from falcon.routing.compiled import _FIELD_PATTERN as FALCON_FIELD_PATTERN
 
-from .._pydantic import ValidationError
+from .._pydantic import BaseModel, ValidationError
 from .._types import ModelType
 from ..response import Response
 from .base import BasePlugin
@@ -227,17 +227,24 @@ class FalconPlugin(BasePlugin):
         func(*args, **kwargs)
 
         if resp and resp.has_model():
-            model = resp.find_model(_resp.status[:3])
-            if model and isinstance(_resp.media, model):
-                _resp.media = _resp.media.dict()
+            model = _resp.media
+            status = int(_resp.status[:3])
+            expect_model = resp.find_model(status)
+            if resp.expect_list_result(status) and isinstance(model, list):
+                _resp.media = [
+                    (entry.dict() if isinstance(entry, BaseModel) else entry)
+                    for entry in model
+                ]
+            elif expect_model and isinstance(_resp.media, expect_model):
+                _resp.media = model.dict()
                 skip_validation = True
 
             if self._data_set_manually(_resp):
                 skip_validation = True
 
-            if model and not skip_validation:
+            if expect_model and not skip_validation:
                 try:
-                    model.parse_obj(_resp.media)
+                    expect_model.parse_obj(_resp.media)
                 except ValidationError as err:
                     resp_validation_error = err
                     _resp.status = HTTP_500
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -231,6 +231,9 @@ class FalconPlugin(BasePlugin):
             status = int(_resp.status[:3])
             expect_model = resp.find_model(status)
             if resp.expect_list_result(status) and isinstance(model, list):
+                expected_list_item_type = resp.get_expected_list_item_type(status)
+                if all(isinstance(entry, expected_list_item_type) for entry in model):
+                    skip_validation = True
                 _resp.media = [
                     (entry.dict() if isinstance(entry, BaseModel) else entry)
                     for entry in model
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -4,6 +4,7 @@ from functools import partial
 from typing import Any, Callable, Dict, List, Mapping, Optional, get_type_hints
 
 from falcon import HTTP_400, HTTP_415, HTTPError
+from falcon import Response as FalconResponse
 from falcon.routing.compiled import _FIELD_PATTERN as FALCON_FIELD_PATTERN
 
 from .._pydantic import BaseModel, ValidationError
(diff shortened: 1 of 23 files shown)


spectree/plugins/falcon_plugin.py
```diff
@@ -227,17 +227,24 @@ def validate(
...
         if resp and resp.has_model():
-            model = resp.find_model(_resp.status[:3])
-            if model and isinstance(_resp.media, model):
-                _resp.media = _resp.media.dict()
+            model = _resp.media
+            status = int(_resp.status[:3])
+            expect_model = resp.find_model(status)
+            if resp.expect_list_result(status) and isinstance(model, list):
+                _resp.media = [
+                    (entry.dict() if isinstance(entry, BaseModel) else entry)
```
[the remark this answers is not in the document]
kemingy: It seems the returned list can contain both BaseModel instances and dictionaries. For example:
```python
return [
    User(user_id="user1").dict(),
    User(user_id="user1"),
]
```
I'm not sure if it should be allowed.

BTW, this implementation breaks the `skip_validation` feature. When the endpoint handler returns a BaseModel instance, we don't need to validate it anymore.

kemingy (commented):
Hi @jean-edouard-boulanger, if I understand it correctly, your expectation is to return a list like `[User(user_id="1"), User(user_id="2")]` directly. I think it's possible but will need to use some tricks.

SpecTree has a construct function called [`gen_list_model`](https://github.com/0b01001001/spectree/blob/bce9963c7e03bcb37b58a23e7ddf1a8ffc3ae37b/spectree/utils.py#L289C12-L289C12). I guess you already notice that during this PR. A possible way is:

```python
UserList = gen_list_model(User)
return UserList(__root__=[User(user_id="1"), User(user_id="2")])
```

spectree/plugins/falcon_plugin.py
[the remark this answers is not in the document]
jean-edouard-boulanger: Hello @kemingy - thanks for taking the time to review this 🙌 

> It seems the returned list can contain both BaseModel instances and dictionaries

Yes absolutely, I've chosen not to carry out any extra validation to keep things as simple as possible. That being said, I also agree that we should perhaps only allow serialized/unserialized models to be returned (which will add additional complexity).

> BTW, this implementation breaks the skip_validation feature

I don't think it breaks it, but I have chosen not to implement it in this first iteration for simplicity.

Let me iterate on those two points (I'll push a second commit for this specifically)

jean-edouard-boulanger:
> Hi @jean-edouard-boulanger, if I understand it correctly, your expectation is to return a list like [User(user_id="1"), User(user_id="2")] directly. I think it's possible but will need to use some tricks.

Yes this is absolutely correct! I don't particularly like the alternative (using `gen_list_model` in the view) and I think it'd be much nicer if spectree allowed returning a list of models directly.

kemingy:
> Yes this is absolutely correct! I don't particularly like the alternative (using `gen_list_model` in the view) and I think it'd be much nicer if spectree allowed returning a list of models directly.

Yes, returning a list directly is more convenient.

As for the `skip_validation` part, if it can return both BaseModel instances and dictionaries, it might be hard to decide whether to validate it or not.

jean-edouard-boulanger:
With the commit I have just pushed, we're getting much closer to your suggestion:

- BaseModel list items are automatically serialized (`.dict()`)
- We skip validation when all list items have the desired/specified type (except with the starlette plugin where this is not possible)
- For all other cases, validation is delegated to Pydantic using the generated model (this is consistent with the existing logic, where Pydantic is always used to validate list results).

Couple of examples:

### Example 1: all item types have the expected/specified type (`User`), validation is skipped

```python
@app.route("/api/example1", methods=["GET"])
@api.validate(resp=Response(HTTP_200=List[User]))
def return_list():
    return [User(id=1), User(id=2)]
```

### Example 2: some items are serialized, some items are not serialized, validation is delegated to Pydantic

```python
@app.route("/api/example2", methods=["GET"])
@api.validate(resp=Response(HTTP_200=List[User]))
def return_list():
    return [User(id=1), {"id": 2}]
```

### Example 3: all items are serialized, validation is delegated to Pydantic (existing use case)

```python
@app.route("/api/example3", methods=["GET"])
@api.validate(resp=Response(HTTP_200=List[User]))
def return_list():
    return [{"id": 1}, {"id": 2}]
```

### Example 4: none of the items have the expected type, validation is delegated to Pydantic (same as example 3)

```python
@app.route("/api/example4", methods=["GET"])
@api.validate(resp=Response(HTTP_200=List[User]))
def return_list():
    return [OtherUserModel(id=1), OtherUserModel(id=2)]
```

spectree/plugins/quart_plugin.py
```diff
@@ -222,6 +222,18 @@ async def validate(
...
+                    skip_validation = True
+                result = (
+                    [
+                        (entry.dict() if isinstance(entry, BaseModel) else entry)
+                        for entry in model
+                    ],
+                    status,
+                    *rest,
+                )
             if expect_model and isinstance(model, expect_model):
```
kemingy, line 237:
```suggestion
                )
            elif expect_model and isinstance(model, expect_model):
```
Align with others.

spectree/plugins/starlette_plugin.py
```diff
@@ -22,7 +22,14 @@ def PydanticResponse(content):
     class _PydanticResponse(JSONResponse):
         def render(self, content) -> bytes:
             self._model_class = content.__class__
```
kemingy, line 24: The returned `List` class will be list, will this affect the correctness?

jean-edouard-boulanger: No it is consistent with the existing implementation (i.e. if you return a list of serialized models,`_model_class` is set to `List`). We could on the other hand extract a bit more metadata and implement validation skipping, but I think it's a bit premature and could do in a followup MR.

kemingy: Sure.

jean-edouard-boulanger:
@kemingy Common validation logic now implemented for both Falcon sync/async implementations (see `FalconPlugin.response_validation`). Tests were updated accordingly.

kemingy (approved):

kemingy:
> Btw, I have noticed that returning Pydantic root models from views is also not supported. Would you be interested in a fix for this too?

I'll take #329. You can fix the views if you're interested. Thanks.

PR #341, opened 2023-09-21
Fix Flask status code processing after release 1.2.2

AndreiDrang:
Following the release of version 1.2.2, the server response's status code generated through "make_response" is no longer being processed accurately. Instead, it is being overwritten with the status code 200.

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -207,7 +207,7 @@ class FlaskPlugin(BasePlugin):
             else:
                 response_payload = result[0]
         else:
-            response_payload = result
+            response_payload, status = result, result.status_code
 
         try:
             response_validation_result = validate_response(
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -206,8 +206,10 @@ class FlaskPlugin(BasePlugin):
                 response_payload, status, *rest = result
             else:
                 response_payload = result[0]
-        else:
+        elif isinstance(result, flask.Response):
             response_payload, status = result, result.status_code
+        else:
+            response_payload = result
 
         try:
             response_validation_result = validate_response(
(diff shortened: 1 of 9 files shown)


AndreiDrang:
Demo example:

```python3
from uuid import uuid4
from datetime import timedelta

from flask import Flask, make_response
from pydantic import Field, BaseModel, constr
from spectree import Response, SpecTree

AUTH_COOKIES_CONFIG = {
    "domain": None,
    "secure": False,
    "httponly": False,
    "samesite": None,
    "max_age": timedelta(days=10),
}


class Profile(BaseModel):
    name: constr(min_length=2, max_length=40)  # constrained str
    age: int = Field(..., gt=0, lt=150, description="user age(Human)")

    class Config:
        schema_extra = {
            # provide an example
            "example": {
                "name": "very_important_user",
                "age": 42,
            }
        }


class Message(BaseModel):
    text: str


app = Flask(__name__)
spec = SpecTree(path="docs", backend_name="flask")


@app.route("/api/user", methods=["POST"])
@spec.validate(json=Profile, resp=Response(HTTP_201=Message, HTTP_403=None), tags=["api"])
def user_profile():
    """
    verify user profile (summary of this endpoint)

    user's name, user's age, ... (long description)
    """
    response = make_response(Message(text="it works!").dict(), 201)
    response.set_cookie(
        key="auth_token",
        value=str(uuid4()),
        **AUTH_COOKIES_CONFIG,
    )
    return response


if __name__ == "__main__":
    spec.register(app)  # if you don't register in api init step
    app.run(port=8000)
```

kemingy (approved):

I will figure out the pypy test failure later.

AndreiDrang:
@kemingy 
Hi
I think we need rollback to 1.2.1 version or smth like this. Cause after 1.2.2 release `make_response` additional logic(like set_cookies\headers and etc.) stop working and need some time to fix it.

kemingy:
> @kemingy Hi I think we need rollback to 1.2.1 version or smth like this. Cause after 1.2.2 release `make_response` additional logic(like set_cookies\headers and etc.) stop working and need some time to fix it.

cc @jean-edouard-boulanger

Let me figure out. Thanks for your feedback.

PR #356, opened 2023-10-26
Add returning result in falcon decorator

BlackLIonz:


spectree/plugins/falcon_plugin.py
```diff
@@ -224,7 +224,7 @@ def validate(
         if req_validation_error:
             return
 
-        func(*args, **kwargs)
+        result = func(*args, **kwargs)
```
kemingy, line 227: Can you explain the reason for this return value?

As far as I know, `falcon` uses `resp` to get the result instead of the `return`.

BlackLIonz: Yes it is, but for custom handlers and/or overrides becomes a bit painful, because of losing warped method's result.

kemingy: I guess you can still get the result from `resp.media` or something like that?

BlackLIonz: It's not only about a resp.media, it's also about not losing return values because of @validate. In my special case I use return in falcon methods and I need to copy paste decorator to not losing return value.

kemingy (approved):

PR #373, opened 2024-06-18
Keep track of resp_validation_error in flask plugin

### LINKED ISSUE #374: bug: resp_validation_error is not passed in after handler of flask plugin
### Describe the bug

In flask plugin when there is a validation error it's not passed in the after handler

### To Reproduce

When using spectree with a flask app and there is a ValidationError and a custom after handler, the handler is not called with resp_validation_error. The value of resp_validation_error is always None

### Expected behavior

_No response_

### The spectree version

1.2.9

### Additional context

_No response_

nikos-hj:
This needs to be passed on to the after handler. 

Closes #374


kemingy (approved):

PR #378, opened 2024-10-29
Fix for openapi.json location if another root directory

Vlczech:
(for example /foo/bar -> /foo/bar/apidoc/openapi.json instead of /apidoc/openapi.json)

### DIFF
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -79,7 +79,7 @@ class FalconPlugin(BasePlugin):
                 f"/{self.config.path}/{ui}",
                 self.DOC_PAGE_ROUTE_CLASS(
                     self.config.page_templates[ui],
-                    spec_url=self.config.spec_url,
+                    spec_url=self.config.filename,
                     spec_path=self.config.path,
                     **self.config.swagger_oauth2_config(),
                 ),
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,9 @@ falcon = [
 flask = [
     "flask",
 ]
+quart = [
+    "quart"
+]
 starlette = [
     "starlette[full]",
 ]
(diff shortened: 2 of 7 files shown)


spectree/plugins/falcon_plugin.py
```diff
@@ -79,7 +79,7 @@ def register_route(self, app: Any):
                 f"/{self.config.path}/{ui}",
                 self.DOC_PAGE_ROUTE_CLASS(
                     self.config.page_templates[ui],
-                    spec_url=self.config.spec_url,
+                    spec_url=self.config.filename,
```
kemingy, line 82: `spec_url` is `f"/{self.path}/{self.filename}"`, I don't get why it should only contain the filename. Can you explain in detail?

Vlczech: As I wrote to comment - absolute address is /apidoc/openapi.json ; so for example for:
* http://foo.bar => http://foo.bar/apidoc.openapi.json
* http://foo.bar/foo/bar/ => http://foo.bar/apidoc/openapi.json

But if there is some subdirectory for API, for example if multiple APIs is available, then absolute address is not good solution. For example base root dirs (basepaths) for APIs:
* http://foo.bar/foo/bar/ => http://foo.bar/apidoc/openapi.json ====\  **conflict!!!**
* http://foo.bar/bar/foo/ => http://foo.bar/apidoc/openapi.json ====/
Right paths should be:
* http://foo.bar/foo/bar/ => http://foo.bar/foo/bar/apidoc/openapi.json
* http://foo.bar/bar/foo/ => http://foo.bar/bar/foo/apidoc/openapi.json

...If you have set some root directory for API, swagger and redoc try to find file absolutely on /apidoc/openapi.json (not relatively) so it show errors only. It is needed to search openapi.json in set subdirectory.

kemingy: Got it. Thanks.

kemingy (approved):

PR #393, opened 2024-12-12
Maintain pydantic.v1 model compatibility while using pydantic v2

voidnologo:
We have pydantic V2 installed in our environemnt, but have been using `from pydantic.v1 import ...`.  With the new changes to support pydantic V2 we are wanting to upgrade Spectree to latest and begin writing new serializers, however, we are not ready to update all of our current models to  be pydantic v2 BaseModels.

Currently if a `ValidationError` is raised, the falcon plug in does not catch it, since `ValidationError` in the plugin is coming from `pydantic`, not `pydantic.v1`.

Since you are aliasing the `pydantic.v1.ValidationError as InternalValidationError`, catching it as well allows us to continue to support our v1 models while we migrate to v2 models.

Looking at the codebase, this looks like a similar issue in the other plugins.  Let me know if you would like me to submit this update for them as well.  I'm not sure where you would like a test for this in the project.

### DIFF
--- a/tests/test_plugin_falcon.py
+++ b/tests/test_plugin_falcon.py
@@ -260,22 +260,22 @@ class CompatibilityView:
     name = "validation works for both pydantic v1 and v2 models simultaneously"
 
     class V1(InternalBaseModel):
-        value: str
+        value: int
 
     class V2(BaseModel):
-        value: str
+        value: int
 
     @api.validate(
         resp=Response(HTTP_200=Resp),
     )
     def on_post_v1(self, req, resp, json: V1):
-        resp.media = "V1"
+        resp.media = Resp(name="falcon v1", score=[1, 2, 3])
 
     @api.validate(
         resp=Response(HTTP_200=Resp),
     )
     def on_post_v2(self, req, resp, json: V2):
-        resp.media = "V2"
+        resp.media = Resp(name="falcon v2", score=[1, 2, 3])
 
 
 app = App()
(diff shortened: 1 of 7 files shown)


voidnologo:
Sorry, it looks like my linter reformatted the file.  Fixing.

timschreiner (approved):
I think I ran into this issue last week and it was making me crazy! good find

joephayes (approved):

kemingy (commented):
LGTM. It would be great if you could also fix other plugins in the same way. 

Adding a test case to verify that v1 validation error can also be caught when using v2 will guarantee that this feature won't be broken in the future.

voidnologo:
When I run tests, it is running against the current published version of spectree (1.4.1).  I bumped the version to 1.4.2 in pyproject.toml.  How do I run tests against the updated code instead?

kemingy:
> When I run tests, it is running against the current published version of spectree (1.4.1). I bumped the version to 1.4.2 in pyproject.toml. How do I run tests against the updated code instead?
> 
> At least its failing in the way we are seeing in our project.

The test will install the current code: https://github.com/0b01001001/spectree/blob/95aff9d9151e00324db865547a44d61517938147/.github/workflows/pythonpackage.yml#L49

So it should be the latest commit in this PR.

One thing to note is that you may need to run `pytest --snapshot-update` if the snapshot test failed due to some changes to the test app API.

tests/test_plugin_falcon.py
```diff
@@ -545,3 +550,40 @@ def test_falcon_optional_alias_response(client):
     )
     assert resp.status_code == 200, resp.json
     assert resp.json == {"schema": "test"}, resp.json
+
+
+@pytest.mark.skipif(not PYDANTIC2, reason="only matters if using both model types")
+def test_falcon_validate_both_v1_and_v2_validation_errors(client):
+    class CompatibilityView:
```
voidnologo, line 557: Inlined this view because adding it to the test app then causes issue with the [generated spec tests](https://github.com/0b01001001/spectree/blob/main/tests/test_plugin.py#L25).

kemingy (approved):

PR #395, opened 2024-12-30
Migrate deprecated Pydantic function

mmathys:
Pydantic's `.schema()` function is deprecated. This throws warnings and will eventually break when Pydantic v3.0 is used with Spectree.

This PR fixes this issue.

Thanks!

### DIFF
--- a/spectree/utils.py
+++ b/spectree/utils.py
@@ -265,7 +265,7 @@ def get_model_schema(
     assert is_pydantic_model(model), f"{model} is not a pydantic model"
 
     nested_key = nested_naming_strategy(naming_strategy(model), "{model}")
-    return model.schema(ref_template=f"#/components/schemas/{nested_key}")
+    return model.model_json_schema(ref_template=f"#/components/schemas/{nested_key}")
 
 
 def get_security(security: Union[None, Mapping, Sequence[Any]]) -> List[Any]:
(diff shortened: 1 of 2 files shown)


mmathys:
However I can't seem to figure out why the tests fail

kemingy:
There are still some legacy pydantic v1 interfaces used internally. To eliminate all the deprecation warnings, we will need to provide another abstraction layer to hide the diff.

The test failed because we tested against both v1&v2.

mmathys:
Thanks @kemingy. I'm not that familiar with the codebase and cannot invest time to this solution. Looking forward to the fix!

kedvall:
If you don't mind keeping this PR open I can try to take a look in the next few days. This was also something I would like to get fixed.

kedvall:
@kemingy I also noticed there's some handling for automatic root model generation in https://github.com/0b01001001/spectree/blob/main/spectree/_pydantic.py#L59. 

I assume this is because a model is required by Response so non-model return types need to be wrapped. Is there a reason this is only done for lists and not done recursively? Would you mind if I add recursive handling and support for dict?

kedvall:
Okay, I'll try to take a look. I think it would be nice to support automatic RootModel conversion for supported response types (i.e. `str`, `bytes`, `dict`, `list` for Flask: https://flask.palletsprojects.com/en/stable/api/#flask.Flask.make_response)

kemingy:
Fixed in https://github.com/0b01001001/spectree/pull/455

PR #408, opened 2025-04-09
fix: conversion to dictionary depending on pydantic version

nightblure:
hi!

having pydantic v2 I wouldn't want to call the `dict` method for at least two reasons:
1. its **deprecated** for v2
2. I could have an overridden implementation of the `dict` method and after moving to v2 it would be logical to **rename** it to `model_dump`. Without this change I'll have to debug the spectree code before I understood that my **method with the new name won't be called**

---

p.s. `noqa: SIM108` it's a matter of taste, I can change it to a ternary operator as the linter suggests, if you want that too


kemingy (approved):

nightblure:
Thanks for such a quick review!

PR #412, opened 2025-04-27
perf (flask): get annotations when no validation error and use cache for type hints

nightblure:
it seems we calculated the annotations anyway, even when there was a validation error - its fixed

and decided to use caching to obtain type hints to improve performance


kemingy (approved):

kemingy:

PR #424, opened 2025-06-25
fix: tuple response handling for flask

smeng9:
Current code has the following bug:

If user's function returns a tuple like ("Success", 200) in result = func(*args, **kwargs), this result tuple will be passed into the flask's make_response function which will throw error

We should pass the payload from payload, status, additional_headers = flask_response_unpack(result) into the make_response function

### DIFF
--- a/spectree/plugins/flask_plugin.py
+++ b/spectree/plugins/flask_plugin.py
@@ -2,6 +2,7 @@ from typing import Any, Callable, Mapping, Optional, Tuple
 
 import flask
 from flask import Blueprint, abort, current_app, jsonify, make_response, request
+from werkzeug.datastructures import Headers
 from werkzeug.routing import parse_converter_args
 
 from spectree._pydantic import (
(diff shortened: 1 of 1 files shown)


kemingy:
Hi @smeng9, do you have a minimal reproducible example for this bug?

smeng9:
Hi @kemingy ,

Here is the minimal reproducible example

from flask import Flask, Response
from spectree import SpecTree, Response as SpecResponse

# Define the request model
class InputPortConfigBody(BaseModel):
    port: int

# Initialize Flask and Spectree
app = Flask(__name__)
spec = SpecTree("flask")

# Register the route and Spectree validation
@app.post("/air-intake-port")
@spec.validate()
def set_air_intake_port(json: InputPortConfigBody) -> tuple[str | Response, int]:
    return f"port {json.port} is set as air intake port", 200

# Register documentation routes (optional)
spec.register(app)

# Entry point
if __name__ == "__main__":
    app.run(debug=True)

smeng9:
Hi @kemingy 

Mypy typing deduction seems have some bug. If I reassign values a, b, c = a.a, a.b, a.c, before a gets assigned, a.b and a.c should still inherit the typing from the original an object. However mypy thinks there are type pollutions before it is assigned.

What is your opinion on this? Shall we disable the mypy checking for this specific line or do you have better ideas to implement this?

smeng9:
> Hi @smeng9,thanks for your feedback.
> 
> I found a solution to the problem. But since you already have the PR here, do you mind me updating on your commits?

Go ahead update my commits

kemingy (approved):
Thanks.

Issue #430, opened 2025-08-21
bug: Falcon demo file upload - AttributeError: 'bytes' object has no attribute 'filename'


tfluehmann:
### Describe the bug

Hello
We are currently implementing the file upload feature from Spectree in conjunction with Falcon 4.x or 3.x. During our testing, we observed that the BaseFile object does not contain a filename when used with Falcon. However, the Falcon demo indicates that a filename should be present. This issue has been tested with both Pydantic versions 2.11 and 1.10.

I am not exactly sure what needs to be done to fix the problem.

### To Reproduce

pyproject.toml
```TOML
[tool.poetry]
name = "spectree-test"
version = "1.0.0"
description = ""
package-mode = false

[tool.poetry.dependencies]
python = ">=3.11,<3.13"
environs = "^11.0.0"
pydantic = "^2.0.0"
falcon = "^4.0.0"
spectree = "^1.4.12"
```

After sending this command:
```BASH
poetry lock
poetry install
dd if=/dev/urandom of=example.bin bs=1M count=1
curl -X POST -F "file=@example.bin" -F "uid=bllaa" http://localhost:8000/api/file_upload
```

Result:
```bash
$ /home/tobias/.cache/pypoetry/virtualenvs/spectree-test-zzOQjvVq-py3.12/bin/python3 falcon_demo.py 
2025-08-21 15:08:42 [FALCON] [ERROR] POST /api/file_upload => Traceback (most recent call last):
  File "/home/tobias/.cache/pypoetry/virtualenvs/spectree-test-zzOQjvVq-py3.12/lib/python3.12/site-packages/falcon/app.py", line 425, in __call__
    responder(req, resp, **params)
  File "/home/tobias/.cache/pypoetry/virtualenvs/spectree-test-zzOQjvVq-py3.12/lib/python3.12/site-packages/spectree/spec.py", line 192, in sync_validate
    return self.backend.validate(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tobias/.cache/pypoetry/virtualenvs/spectree-test-zzOQjvVq-py3.12/lib/python3.12/site-packages/spectree/plugins/falcon_plugin.py", line 281, in validate
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/tobias/projects/spectree/examples/falcon_demo.py", line 133, in on_post
    resp.media = {"filename": file.filename, "type": file.type}
                              ^^^^^^^^^^^^^
[...]

PR #434, opened 2025-08-22
fix: falcon form with `application/x-www-form-urlencoded`

kemingy:
- fix https://github.com/0b01001001/spectree/issues/430#issuecomment-3213660670
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type#directives

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "spectree"
-version = "1.5.1"
+version = "1.5.2"
 dynamic = []
 description = "Generate OpenAPI document and validate request & response with Python annotations."
 readme = "README.md"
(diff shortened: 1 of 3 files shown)


tfluehmann:
Works for me! 
```bash
>>> import requests
>>> url = 'http://localhost:8000/api/file_upload'
>>> files = {'file': open('example.bin', 'rb')}
>>> data = {'uid': 'bla'}
>>> response = requests.post(url, files=files, data=data)
>>> print(response.text)
{"filename": "example.bin", "type": "text/plain", "uid": "bla"}
```

```python
    @spec.validate(resp=Response(HTTP_200=FileResp), tags=["file-upload"])
    def on_post(self, req, resp, form: File):
        """
        post multipart/form-data demo

        demo for 'form'
        """
        file: BodyPart = form.file
        response = {}
        if file:
            response = {"filename": file.secure_filename, "type": file.content_type}
        if form.uid:
            response.update({"uid": form.uid})
        resp.media = response
```


```

Issue #249, opened 2022-08-04
improve falcon form data read part


kemingy:
> @yedpodtrzitko It depends what you are trying to achieve. Do you want to buffer whole files in memory? Or do you want to spool them to temp files like some other frameworks do?

> Falcon even puts a cap on a maximum amount of data that can be referenced this way ([`await part.get_data()`](https://falcon.readthedocs.io/en/stable/api/multipart.html#falcon.media.multipart.BodyPart.get_data)) in order to avoid surprises such as running out of memory.

> Use `await part.stream.read()` to read the whole part as a bytestring, or [`await part.stream.pipe(async_file)`](https://falcon.readthedocs.io/en/stable/api/multipart.html#multipart-forms), or read by chunks, and store the result somewhere. You'll probably need to introduce some new object type to hold these attributes.

_Originally posted by @vytas7 in https://github.com/0b01001001/spectree/pull/225#discussion_r936042043_

vytas7:
@kemingy I'm happy to help or at least advise, but could you shed some light on Spectree's design wrt my questions above. What type of data from a multipart form is typically sent to the validator?

kemingy:
> @kemingy I'm happy to help or at least advise, but could you shed some light on Spectree's design wrt my questions above. What type of data from a multipart form is typically sent to the validator?

Thanks! Here is the current implementation for the form file type:
https://github.com/0b01001001/spectree/blob/3deaf0009ccdbb914aed6c0d3994edc5130e2dae/spectree/models.py#L173-L192

In my option, I would prefer to leave the large form body to falcon (since it's already there). I feel SpecTree doesn't really "validate" the body. I wonder if SpecTree should provide a struct that when users try to read it, users actually read from falcon. Buffering it to a temp file also looks good.

vytas7:
I see. Falcon offers parsing on-the-fly without buffering anything anywhere, so spooling to a temp file should be built in the Spectree plugin, if this is desired.

kemingy:
The current validation doesn't touch the multiple files. The attribute marked as `file: BaseFile` will be assigned as the target framework file object. In Falcon, it's the `BodyPart`.

Ideally, I hope we can pass the files to the user without buffering in memory. This might not be possible since the `multipart/form-data` can contain some other key-value pairs, which need to be validated.

If we cannot do that, we might need the [SpooledTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile).

PR #435, opened 2025-08-23
fix: keep the falcon file stream in memory

kemingy:
Address the comments from https://github.com/0b01001001/spectree/pull/434/files#r2293604269

1. remove the try...except, it seems those exceptions are not meant to be caught in this layer
2. replace `data` with `get_data()`
3. test both `application/x-www-form-urlencoded` & `multipart/form-data`
4. wrap the stream (this is a bad idea for large files)

It seems the falcon internal stream only lives in the `request.get_media().__iter__`. After that, it will be empty. But we have to go through the `__iter__` here.

Not sure if there are any better ways.

### DIFF
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -215,15 +215,13 @@ class FalconPlugin(BasePlugin):
         if cookies:
             req.context.cookies = cookies.parse_obj(req.cookies)
         if json:
-            media = req.media
-            req.context.json = json.parse_obj(media)
+            req.context.json = json.parse_obj(req.get_media())
         if form and req.content_type:
-            form_data = req.get_media()
+            req_form = {}
             if req.content_type == "application/x-www-form-urlencoded":
-                req_form = form_data
+                req_form = req.get_media()
             elif req.content_type.startswith("multipart/form-data"):
-                req_form = {}
-                for part in form_data:
+                for part in req.get_media():
                     if part.filename is None:
                         req_form[part.name] = part.get_data()
                     else:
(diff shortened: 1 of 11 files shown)


spectree/plugins/falcon_plugin.py
```diff
@@ -185,26 +215,23 @@ def validate_request(self, req: FalconRequest, query, json, form, headers, cooki
...
-                media = req.media
-            except HTTPError as err:
-                if err.status not in self.FALCON_MEDIA_ERROR_CODE:
-                    raise
-                media = None
+            media = req.media
             req.context.json = json.parse_obj(media)
         if form and req.content_type:
+            form_data = req.get_media()
             if req.content_type == "application/x-www-form-urlencoded":
```
tfluehmann, line 222: I found one case in my tests where we land in here with a json request:
image
In this case `req_form` is unassigned.
It might be good to assign it in the beginning of the if statement or add it in else.

kemingy: Right, some cases don't need the media.

kemingy:
> @kemingy thanks a lot for your work. Can we release in a bugfix release for #[22a39a1](https://github.com/0b01001001/spectree/pull/435/commits/22a39a1479f2ee31ad7a187fff9e0115dcb1cb9c) soon if the whole change takes longer?

I think we can merge this one for now. A better solution can be discussed in #249.

spectree/plugins/falcon_plugin.py
```diff
@@ -22,6 +31,28 @@
 from spectree.utils import cached_type_hints
 
 
+class StreamWrapper:
+    def __init__(self, stream):
+        self._buf = CachedFile()
```
vytas7, line 36: I'm a bit leery of replacing a body part's `.stream` attribute, but I guess we don't offer easy workarounds for that atm...

How are these files cleaned up?

As I understand at least on Linux, they are created using the `O_TMPFILE` flag which makes them go away when the handle is closed. However, that only happens when the part object is garbage collected, since we are not closing anything explicitly.

Maybe it is good enough? :thinking:

kemingy: Yeah. I have to replace the `.stream` and provide a similar interface.

The files are cleaned by garbage collection. I didn't find a way to close it explicitly.

Issue #437, opened 2025-08-26
bug: json_compatible_deepcopy breaks compatibility with enum types


tfluehmann:
### Describe the bug

We found a problem with spectree versions after https://github.com/0b01001001/spectree/commit/d33573a152b45a378b2e2402e37bf8a855ba361f

When using pydantic v1 it raises the following error:
`TypeError: Unsupported type for this custom deepcopy: <enum 'CustomEnum'>`

It happens when the model uses some slightly more sophisticated validations like confrozenset, see in the reproduce steps.

```bash
Traceback (most recent call last):
  File "/home/tobias/projects/spectree/examples/falcon_demo.py", line 155, in <module>
    class EnumTest:
  File "/home/tobias/projects/spectree/examples/falcon_demo.py", line 161, in EnumTest
    def on_get(self, req, resp, form: CustomType):
  File "/home/tobias/projects/spectree/spectree/spec.py", line 246, in decorate_validation
    model_key = self._add_model(model=model)
  File "/home/tobias/projects/spectree/spectree/spec.py", line 277, in _add_model
    self.models[model_key] = json_compatible_deepcopy(
  File "/home/tobias/projects/spectree/spectree/utils.py", line 389, in json_compatible_deepcopy
    return naive_deepcopy(obj)
  File "/home/tobias/projects/spectree/spectree/utils.py", line 374, in naive_deepcopy
    res = {
  File "/home/tobias/projects/spectree/spectree/utils.py", line 375, in <dictcomp>
    naive_deepcopy(key): naive_deepcopy(value) for key, value in obj.items()
  File "/home/tobias/projects/spectree/spectree/utils.py", line 374, in naive_deepcopy
    res = {
  File "/home/tobias/projects/spectree/spectree/utils.py", line 375, in <dictcomp>
    naive_deepcopy(key): naive_deepcopy(value) for key, value in obj.items()
  File "/home/tobias/projects/spectree/spectree/utils.py", line 374, in naive_deepcopy
    res = {
  File "/home/tobias/projects/spectree/spectree/utils.py", line 375, in <dictcomp>
    naive_deepcopy(key): naive_deepcopy(value) for key, value in obj.items()
  File "/home/tobias/projects/spectree/spectree/utils.py", line 378, in naive_deepcopy
    res = [naive_deepcopy(item) for item in obj]
  File "/home/tobias/projects/spectree/spectree/utils.py", line 378, in <listcomp>
    res = [naive_deepcopy(item) for item in obj]
[...]

PR #438, opened 2025-08-26
fix: support enum value in naive deep copy

kemingy:
- fix https://github.com/0b01001001/spectree/issues/437


tfluehmann:
nice thanks, looks legit!

PR #451, opened 2025-09-18
fix: use pipe for falcon stream, run async write in the event loop

### LINKED ISSUE #249: improve falcon form data read part

### LINKED ISSUE #440: use async io for falcon async multipart file read/write
Shouldn't this be scheduled in a thread? It might potentially block the whole app.

_Originally posted by @vytas7 in https://github.com/0b01001001/spectree/pull/435#discussion_r2304457050_

### LINKED ISSUE #448: bug: Falcon: multipart body part shouldn't be read into memory
### Describe the bug

Currently, the non-standard form stream shim is populated via [`part.stream.read()`](https://github.com/0b01001001/spectree/blob/main/spectree/plugins/falcon_plugin.py#L235), which can read an unbound amount of data into the memory.

A better course of action would be to open the spooled file (or another file-like destination), and use `part.stream.pipe(...)` as in the example on the top of this page: [Multipart Forms](https://falcon.readthedocs.io/en/stable/api/multipart.html).

The same solution works for the `async` flavour, however, there piping needs to be `await`-ed too, so it doesn't work with a sync file obj out of the box. An instance of, e.g., `aiofiles` would work. So I guess the `async` part can be coordinated with #440.

### To Reproduce

Upload a form containing a large file(s). Watch memory usage to temporarily increase by this amount.

kemingy:
- fix #448 
- fix #440 
- fix #249

### DIFF
--- a/spectree/plugins/falcon_plugin.py
+++ b/spectree/plugins/falcon_plugin.py
@@ -16,7 +16,9 @@ from falcon import MEDIA_JSON, http_status_to_code
 from falcon import Request as FalconRequest
 from falcon import Response as FalconResponse
 from falcon.asgi import Request as FalconASGIRequest
+from falcon.asgi.reader import BufferedReader as ASGIBufferedReader
 from falcon.routing.compiled import _FIELD_PATTERN as FALCON_FIELD_PATTERN
+from falcon.util.reader import BufferedReader
 
 from spectree._pydantic import (
     InternalValidationError,
@@ -32,9 +34,9 @@ from spectree.utils import cached_type_hints
 
 
 class StreamWrapper:
-    def __init__(self, stream):
+    def __init__(self, stream: BufferedReader):
         self._buf = CachedFile()
-        self._buf.write(stream)
+        stream.pipe(self._buf)
         self._buf.seek(0)
 
     def read(self, size: Optional[int] = -1, /) -> bytes:
@@ -46,6 +48,15 @@ class StreamWrapper:
 
 
 class AsyncStreamWrapper(StreamWrapper):
+    def __init__(self):
+        self._buf = CachedFile()
+
+    @classmethod
+    async def from_stream(cls, stream: ASGIBufferedReader):
+        obj = cls()
+        await stream.pipe(obj._buf)
+        obj._buf.seek(0)
+
     async def read(self, size: Optional[int] = -1, /) -> bytes:  # type: ignore[override]
         return super().read(size)
 
@@ -232,7 +243,7 @@ class FalconPlugin(BasePlugin):
                         req_form[part.name] = part
                         # try to consume the file data, otherwise it will be lost
                         # this is hacky since it changed the underlying stream type
-                        part.stream = StreamWrapper(part.stream.read())
+                        part.stream = StreamWrapper(part.stream)
             req.context.form = form.parse_obj(req_form)
 
     def validate_response(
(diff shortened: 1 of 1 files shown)


kemingy:
cc @vytas7

I saw the falcon sync BufferedReader.read will treat `-1 or None` as max size, but the async BufferedReader.read will treat `None` as 0, is that correct?
- sync https://github.com/falconry/falcon/blob/305a0b7f58a6cbb002d68d20d0e96098492747c4/falcon/util/reader.py#L123
- async https://github.com/falconry/falcon/blob/305a0b7f58a6cbb002d68d20d0e96098492747c4/falcon/asgi/reader.py#L303

vytas7:
> cc @vytas7
> 
> I saw the falcon sync BufferedReader.read will treat `-1 or None` as max size, but the async BufferedReader.read will treat `None` as 0, is that correct?
> 
> * sync https://github.com/falconry/falcon/blob/305a0b7f58a6cbb002d68d20d0e96098492747c4/falcon/util/reader.py#L123
> * async https://github.com/falconry/falcon/blob/305a0b7f58a6cbb002d68d20d0e96098492747c4/falcon/asgi/reader.py#L303

Hi, no, that's probably a bug in Falcon if it behaves that way... I'll check.

spectree/plugins/falcon_plugin.py
```diff
@@ -32,20 +35,34 @@
...
 class AsyncStreamWrapper(StreamWrapper):
+    def __init__(self):
+        self._buf = CachedFile()
+
+    @classmethod
+    async def from_stream(cls, stream: ASGIBufferedReader):
+        obj = cls()
+        loop = asyncio.get_running_loop()
+        # copied from the falcon ASGIBufferedReader pipe implementation
+        async for chunk in stream._iter_with_buffer():
```
vytas7, line 61: Copilot is surprisingly right here 😅

Don't use private interfaces like that, we can break this at any time without knowing or intending it.

Just iterate the stream object itself, as shown here: https://falcon.readthedocs.io/en/stable/api/multipart.html (click on the ASGI tab of the top example):
```python
 # Do something with the uploaded data (file)
        async for chunk in part.stream:
            await feed_data(chunk)
```

kemingy: Fixed. Thanks.

spectree/plugins/falcon_plugin.py
```diff
@@ -32,20 +35,33 @@
...
+    def __init__(self):
+        self._buf = CachedFile()
+
+    @classmethod
+    async def from_stream(cls, stream: ASGIBufferedReader):
+        obj = cls()
+        loop = asyncio.get_running_loop()
+        async for chunk in stream:
+            await loop.run_in_executor(None, obj._buf.write, chunk)
+        obj._buf.seek(0)
```
vytas7, line 62: Probably more a theoretical edge case, but seeking should also be run in an executor.

It can block, especially on network file systems, or essentially any kind of storage if the file is large enough so that the beginning of it is not cached anymore, although I guess practically on an SSD the corresponding syscall will be blazing fast.

At least `aiofiles` seems to delegate this to an executor as well: https://github.com/Tinche/aiofiles/blob/main/src/aiofiles/threadpool/binary.py#L42.

vytas7 (approved):

PR #493, opened 2026-05-30
fix: ignore unresolvable return annotation in annotations mode (#312)

### LINKED ISSUE #312: bug: annotations NameError issue caused by python bug
### Describe the bug

Hey :) This isn't strictly a bug in Spectree but I wanted to raise it nonetheless (see below as to why).

The combination of Spectree, annotations mode and Flask's ResponseReturnValue does not work:

```
   File "/path/to/viewfuncs.py", line 28, in <module>
     @api.validate(
      ^^^^^^^^^^^^^
   File "/usr/local/lib/python3.11/site-packages/spectree/spec.py", line 221, in decorate_validation
     annotations = get_type_hints(func)
                   ^^^^^^^^^^^^^^^^^^^^
   File "/usr/local/lib/python3.11/typing.py", line 2361, in get_type_hints
     hints[name] = _eval_type(value, globalns, localns)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/usr/local/lib/python3.11/typing.py", line 380, in _eval_type
     ev_args = tuple(_eval_type(a, globalns, localns, recursive_guard) for a in t.__args__)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/usr/local/lib/python3.11/typing.py", line 380, in <genexpr>
     ev_args = tuple(_eval_type(a, globalns, localns, recursive_guard) for a in t.__args__)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/usr/local/lib/python3.11/typing.py", line 366, in _eval_type
     return t._evaluate(globalns, localns, recursive_guard)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/usr/local/lib/python3.11/typing.py", line 864, in _evaluate
     eval(self.__forward_code__, globalns, localns),
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "<string>", line 1, in <module>
 NameError: name 'Response' is not defined
```

The cause is this bug in Python: https://bugs.python.org/issue43463

The tl;dr is: spectree calls `get_type_hints`, that in turn evaluates the function's return type, in this case Flask's `ResponseReturnValue`, this refers to a `Response` class, but it is done so via a string:

```
# The possible types that are directly convertible or are a Response object.
ResponseValue = t.Union[
    "Response",
    str,
    bytes,
    t.List[t.Any],
    # Only dict is actually accepted, but Mapping allows for TypedDict.
    t.Mapping[str, t.Any],
    t.Iterator[str],
    t.Iterator[bytes],
]
```

The `Response` class is not imported at runtime, it's behind a `TYPE_CHECKING` guard:

```
if t.TYPE_CHECKING:  # pragma: no cover
    from werkzeug.wrappers import Response  # noqa: F401
```

Thus, Spectree's annotations feature won't work if the function has a type hint where one of the constituent parts is only imported behind `TYPE_CHECKING`. The python bug doesn't appear as if it will have a resolution any time soon, sadly.

I am not very experienced with python typing, but, I wonder if spectree should be using `get_type_hints`. Underneath `get_type_hints` uses `.__annotations__` which appears to already have the types spectree needs. Thus I wonder if spectree could use that rather than `get_type_hints`, thus sidestepping the issue. WDYT?

_p.s many thanks for creating and maintaining spectree <3_

### To Reproduce

```
from flask.typing import ResponseReturnValue

# opt in into annotations feature
spec = SpecTree("flask", annotations=True)


@app.route("/api/user", methods=["POST"])
@spec.validate(resp=Response(HTTP_200=Message, HTTP_403=None), tags=["api"])
def user_profile(json: Profile) -> ResponseReturnValue:
    """
    verify user profile (summary of this endpoint)

    user's name, user's age, ... (long description)
    """
    print(json)  # or `request.json`
    return jsonify(text="it works")  # or `Message(text='it works')`
```

### Expected behavior

Spectree/Flask starts up

### The spectree version

- Spectree version 1.1.2
- `Linux efeb38394e42 6.3.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 11 May 2023 16:40:42 +0000 x86_64 GNU/Linux`
- `Python 3.11.3`

### Additional context

_No response_

MohammadYusif:
## Summary

- With `annotations=True`, `SpecTree.validate` resolved a view's type hints
  via `get_type_hints(func)`, which evaluates *every* annotation including
  the return type — but spectree only reads the request-parameter annotations
  (`query`, `json`, `form`, `headers`, `cookies`).
- A return annotation referencing a name imported only under
  `typing.TYPE_CHECKING` (e.g. Flask's `flask.typing.ResponseReturnValue`)
  therefore made decoration crash with `NameError`, even though the return
  type is never used (CPython bpo-43463).
- This fix resolves the parameter annotations without letting an unresolvable
  return annotation break them.

## Changes

- `spectree/utils.py`: add `get_parameter_type_hints(func)`, which temporarily
  drops the unused `return` annotation before calling `get_type_hints`
  (restoring it afterward; `__globals__` untouched so parameter forward refs
  still resolve)
- `spectree/spec.py`: call `get_parameter_type_hints` instead of
  `get_type_hints`; remove the now-unused `get_type_hints` import
- `tests/type_checking_annotation_case.py`: fixture reproducing the crash
- `tests/test_utils.py`: regression test — plain `get_type_hints` raises
  `NameError` on the fixture; the helper returns `{"json": DemoModel}`
  and leaves the annotation intact

Fixes #312

### DIFF
--- a/spectree/utils.py
+++ b/spectree/utils.py
@@ -5,7 +5,7 @@ import re
 from enum import Enum
 from hashlib import sha1
 from math import isinf, isnan
-from types import UnionType
+from types import FunctionType, UnionType
 from typing import (
     Annotated,
     Any,
@@ -327,16 +327,25 @@ def get_parameter_type_hints(func: Callable[..., Any]) -> Mapping[str, Any]:
     need. The function's ``__globals__`` are left untouched so forward
     references in the parameter annotations still resolve correctly.
     """
-    annotations = getattr(func, "__annotations__", None)
-    if not annotations or "return" not in annotations:
+    annotations = getattr(func, "__annotations__", {})
+    if "return" not in annotations:
         return get_type_hints(func)
 
-    return_annotation = annotations["return"]
-    del annotations["return"]
-    try:
-        return get_type_hints(func)
-    finally:
-        annotations["return"] = return_annotation
+    # Build a thread-safe proxy that shares __globals__ with the original but
+    # carries only the parameter annotations (no 'return').  Mutating
+    # func.__annotations__ in place is not thread-safe, so we create a new
+    # FunctionType object instead.  If a *parameter* annotation is
+    # unresolvable, the NameError propagates — spectree genuinely needs those
+    # types.
+    proxy = FunctionType(
+        func.__code__,
+        func.__globals__,
+        func.__name__,
+        func.__defaults__,
+        func.__closure__,
+    )
+    proxy.__annotations__ = {k: v for k, v in annotations.items() if k != "return"}
+    return get_type_hints(proxy)
 
 
 def is_list_item(key: str, model: Optional[ModelClass]) -> bool:
--- a/spectree/utils.py
+++ b/spectree/utils.py
@@ -308,6 +308,37 @@ def get_multidict_items_starlette(
     return res
 
 
+def get_parameter_type_hints(func: Callable[..., Any]) -> Mapping[str, Any]:
+    """Resolve the parameter type hints of ``func``.
+
+    :meth:`SpecTree.validate` only reads the annotations of the request
+    parameters (``query``, ``json``, ``form``, ``headers``, ``cookies``); the
+    return annotation is never used. :func:`typing.get_type_hints` however
+    evaluates *every* annotation on the function, including the return type.
+
+    A return annotation that references a name only imported under
+    ``typing.TYPE_CHECKING`` (e.g. Flask's ``ResponseReturnValue``) therefore
+    makes ``get_type_hints`` raise ``NameError`` even though spectree never
+    needs that annotation. See https://github.com/0b01001001/spectree/issues/312
+    and the underlying CPython limitation https://bugs.python.org/issue43463.
+
+    To stay robust, temporarily drop the return annotation before resolving so
+    an unresolvable return type cannot break the parameter annotations we do
+    need. The function's ``__globals__`` are left untouched so forward
+    references in the parameter annotations still resolve correctly.
+    """
+    annotations = getattr(func, "__annotations__", None)
+    if not annotations or "return" not in annotations:
+        return get_type_hints(func)
+
+    return_annotation = annotations["return"]
+    del annotations["return"]
+    try:
+        return get_type_hints(func)
+    finally:
+        annotations["return"] = return_annotation
+
+
 def is_list_item(key: str, model: Optional[ModelClass]) -> bool:
     """Check if this key is a list item in the model."""
     if model is None:
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -316,6 +316,17 @@ def test_get_parameter_type_hints_unresolvable_return_annotation():
     assert "return" in type_checking_view_func.__annotations__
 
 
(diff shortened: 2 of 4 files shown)


kemingy:
Thanks for the PR. I'm extremely busy recently. Will try to review this on the weekend. Feel free to comment here to remind me if I forget.

spectree/utils.py
```diff
@@ -308,6 +308,37 @@ def get_multidict_items_starlette(
...
+    annotations = getattr(func, "__annotations__", None)
+    if not annotations or "return" not in annotations:
+        return get_type_hints(func)
+
+    return_annotation = annotations["return"]
+    del annotations["return"]
+    try:
+        return get_type_hints(func)
+    finally:
+        annotations["return"] = return_annotation
```
kemingy: My concern is that the parameter types can also behind the `TYPE_CHECKING`. So we'd better check the required:

- json
- form
- query
- headers
- cookies

If any of them cannot be evaluated, there is nothing we can do but raise exceptions.

Also, modifying the `__annotations__` is not thread safe, I would prefer construct a proxy func to evaluate the types.

tests/test_utils.py
```diff
@@ -291,6 +293,29 @@ def test_parse_params_with_route_param_keywords():
...
+    def func(query: DemoQuery, json: DemoModel) -> int:
+        return 0
+
+    hints = get_parameter_type_hints(func)
+    assert hints["query"] is DemoQuery
+    assert hints["json"] is DemoModel
+    assert "return" not in hints
+
+
+def test_get_parameter_type_hints_unresolvable_return_annotation():
```
kemingy, line 306: Should also cover the parameters, not just response types.

MohammadYusif:
Thanks for the feedback! I've pushed an update (4a52684):

- **Thread safety**: replaced the `del` / `finally restore` pattern with a `FunctionType` proxy that shares `__globals__` with the original but carries only the parameter annotations — the original `func.__annotations__` is never touched.
- **Parameter annotation test**: added `test_get_parameter_type_hints_unresolvable_parameter_annotation` which asserts that a `NameError` on an unresolvable parameter annotation propagates (spectree needs those types, so it should raise).

spectree/utils.py
```diff
@@ -308,6 +308,46 @@ def get_multidict_items_starlette(
     return res
 
 
+def get_parameter_type_hints(func: Callable[..., Any]) -> Mapping[str, Any]:
+    """Resolve the parameter type hints of ``func``.
+
+    :meth:`SpecTree.validate` only reads the annotations of the request
+    parameters (``query``, ``json``, ``form``, ``headers``, ``cookies``); the
```
kemingy, line 315: This doesn't reveal in the code.

kemingy:
Fixed in #508. Thanks for your help.

